PR #133, opened 2015-01-03
SOCKS5 support

klondi:
This adds partial SOCKS5 support, i.e. no valid addresses are returned yet and no support for bind, UDP and GSSAPI (I doubt we'll ever have GSSAPI anyways).

It also includes some small cleanups fixing the issues causes by HOST_NAME_MAX


klondi:
This also fixes a strange bug I found where a \0 was attached to the end of SOCKS4a replies.

PR #169, opened 2015-03-27
the binary should be i2pd

guanqun:
Shouldn't the name be `i2pd`? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ On Ubuntu/Debian based
 
 Then, run it:
 
-$ ./i2p
+$ ./i2pd
 
 The client should now reseed by itself.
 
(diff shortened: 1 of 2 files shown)


orignal:
Now now, should be discussed with kytv first.

guanqun:
I'm using `cmake` to compile and the generated binary name is `i2pd`, isn't it?

orignal:
Probably, but plain Makefile still generates i2p

kytvi2p:
It's fine with me. ;)

guanqun:
I can change the plain Makefile so that it generates `i2pd`, how do you think?

orignal:
Again, it must be discused with kytv about his packaging

meeh420:
Kytv said it was fine.

PR #177, opened 2015-04-07
Fixed memory leak: delete -> delete[]

robertfoss:


orignal:
that's not a leak really, because it's array of uint8s.

PR #198, opened 2015-06-06
Fix UPnP for Win32

mlt:
* find_package for headers
* Swap includes order to pass compilation with MSVC 2013
* Enforce SO address resolution checks
* Change SO/DLL name on Windows
* Portable sleep from C++11

### DIFF
--- a/UPnP.cpp
+++ b/UPnP.cpp
@@ -2,15 +2,19 @@
 #include <string>
 #include <thread>
 
-#ifdef _WIN32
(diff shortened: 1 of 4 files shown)


mlt:
Is library/binary separation a WIP? They are quite intertwined, namely UPnP and [ClientContext.h](https://github.com/PurpleI2P/i2pd/blob/master/ClientContext.h#L82).

PR #214, opened 2015-06-20
Missing libdl for UPnP

mlt:
It is definitely missing. I'm not sure if WIN32 shall be used instead of MSVC...


mlt:
apparently github doesn't pick up same name branch after forced push

mlt:
huh! nope, it does :smiley: here we go forced pushes

PR #249, opened 2015-08-11
make http server http/1.1 compliant

majestrate:
It also included an extra null char at the end of the first line, fixed too.


EinMByte:
This will be merged together with pull request #251.

PR #275, opened 2015-09-30
add datagram forwarding to sam

majestrate:


EinMByte:
Seems ok, will merge and maybe make some minor edits soon.
Could be in 0.11 it is works well, or maybe we should wait.

PR #281, opened 2015-10-13
Catch exception in SSUServer that would dump core.

### LINKED ISSUE #280: Uncaught SSUServer exception leads to core dump
```
17:11:57/info - Creating new SSU session to [8irb] 94.114.203.244:37194
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  send_to: Operation not permitted
Aborted (core dumped)
$
```
An iptables line proved to be the culprit (this machine intentionally blocks non-DNS outbound udp):
```
-A OUTPUT ! -o lo -p udp -m udp ! --dport 53 -j REJECT --reject-with icmp-admin-prohibited
```

Simply removing the line is a temporary fix. Sending a PR patch in a few minutes.

anonimal:
Fixes #280 

I'd rather leave this as-is with the assumption that in the future, i2pd will have a "disable outbound UDP" option - and that will eventually lead to some re-writing anyway. If there is a better way to handle this for now, I'm all ears!

### DIFF
--- a/core/RouterInfo.cpp
+++ b/core/RouterInfo.cpp
@@ -426,7 +426,7 @@ namespace data
 
     void RouterInfo::CreateBuffer (const PrivateKeys& privateKeys)
     {
-        m_Timestamp = i2p::util::GetMillisecondsSinceEpoch (); // refresh timstamp
+        m_Timestamp = i2p::util::GetMillisecondsSinceEpoch (); // refresh timestamp
         std::stringstream s;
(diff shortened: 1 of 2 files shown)


EinMByte:
I'll merge this as a temporary solution. As you said, the real solution is to disable outbound UDP when this fails (automatically or through an option).
See #282 for more.

Issue #244, opened 2015-08-02
Use Asio SSL and OpenSSL for HTTPS


EinMByte:
[Boost.Asio](http://www.boost.org/doc/libs/1_58_0/doc/html/boost_asio/overview/ssl.html) allows layering SSL without much changes to existing networking code. I think it's a (very) bad idea to keep using some home-brew SSL implementation.

Issue #279, opened 2015-10-10
Roadmap 2016


anonimal:
The following has moved to https://github.com/monero-project/kovri
=========================================
I know this may be a bit ambitious, but I think that a roadmap would be great. It's also been a subject of inquiry, so I think that having some solid ground would (at the very least) "look good". I also spoke briefly with @fluffypony about an irc meeting at some point, but everyone's schedule, etc...

Thinking out-loud, I know that a few of my personal ambitions are:

1. Squashing odd startup bugs and opening tickets when ready (hangs when SU3 is 404, ~~core dumps~~ (Fixed in #281), etc.)
2. Reworking some signal handling ~~(currently not working in Arch)~~ (most likely not Arch related)
3. Reworking the UI (including options for real-time stats)
4. Much longer term goals include: hidden mode and disabling inbound
5. Oh yeah, and all these open tickets...

I currently lack the competency to go anywhere near 4. or most of 5., but there are plenty of other things to do in the mean time.

I've contributed some work into a VRP at http://trac.i2p2.i2p/wiki/Roadmaps/2h2015 that we could use as a starting point in at least that area. Maybe running with dates/ideas that are in parallel could also be beneficial.

I'm open to all ideas and will be happy (and would prefer) to start sketching things out before tackling my wishlist. By doing so, we'd all be on the same page from the get-go - and I think that would be great.

EinMByte:
IRC meeting is probably a good idea.
There are quite a few open issues at the moment, in particular short-term fixes should be #240, #236, #267, #261, #244,  #234.
The next release (0.11) was planned a few weeks ago, but I'll try to do it soon.

Another very important change we need is peer profiling, this is probably why tunnel build success rates are so low right now with i2pd.

anonimal:
Hi, sorry to ping you all, but I'm wondering if you'd be interested in having an irc meeting to discuss the project since the dust has settled and most of the drama has passed (I hope). Even if the meeting is weeks or months from now, I think it would be great to set a date. @majestrate has done some cool webUI work and I have some thoughts (early work) on an arm-like monitor for i2pd (c++/ncurses, bundled or separate project to be determined) but of course, as @EinMByte pointed out, there are other issues at hand and I hope we can chat about them.

Does November 1st, 8pm UTC work for you?

@meeh420 @klondi @kytvi2p @mlt @fluffypony

mlt:
[...]

PR #283, opened 2015-10-15
Improve HTTP/S header handling, reorganize HTTPS, note on TLS.

anonimal:
If there was a maintenance or development branch I would PR there, but I think this will do for now.

When bug squashing, I was actually quite confused as to what the relationship of HTTPS was to reseeding so I reviewed and found that there wasn't any - aside from making an HTTPS request to a reseed server (please correct me if I'm wrong). I can see the crypto-logic for keeping the request there but implementation-wise, I found it confusing and contrary to what will eventually be needed for a #244 fix.

I think this PR will streamline the HTTP/S request process and leave reseeding to actual reseeding. I didn't bother touching the TLS portion because of #244 (of which I <b>very</b> much agree with you on) so I simply added a comment for clarity.

### DIFF
--- a/client/ClientContext.cpp
+++ b/client/ClientContext.cpp
@@ -71,7 +71,7 @@ namespace client
                 ircPort, localDestination
             );
             ircTunnel->Start ();
-            // TODO: allow muliple tunnels on the same port (but on a different address)
+            // TODO: allow multiple tunnels on the same port (but on a different address)
             m_ClientTunnels.insert(std::make_pair(
                 ircPort, std::unique_ptr<I2PClientTunnel>(ircTunnel)
             ));
--- a/client/Daemon.cpp
+++ b/client/Daemon.cpp
@@ -76,9 +76,11 @@ namespace i2p
             int port = i2p::util::config::GetArg("-port", 0);
             if (port)
                 i2p::context.UpdatePort (port);                 
-            const char * host = i2p::util::config::GetCharArg("-host", "");
-            if (host && host[0])
-                i2p::context.UpdateAddress (boost::asio::ip::address::from_string (host));  
+            const std::string host = i2p::util::config::GetArg("-host", "");
+            // The host option is deprecated, so this was always true.
+            // Because of -Waddress, it's now exposed as the hack it was.
+            if (&host && host[0])
+                i2p::context.UpdateAddress (boost::asio::ip::address::from_string (host));
 
             i2p::context.SetSupportsV6 (i2p::util::config::GetArg("-v6", 0));
             i2p::context.SetFloodfill (i2p::util::config::GetArg("-floodfill", 0));
--- a/core/AddressBook.cpp
+++ b/core/AddressBook.cpp
@@ -487,11 +487,10 @@ namespace client
             {
                 std::stringstream request, response;
                 // standard header
-                request << "GET " << u.path_ << " HTTP/1.1\r\nHost: " << u.host_
-                << "\r\nAccept: */*\r\n" << "User-Agent: Wget/1.11.4\r\n" << "Connection: close\r\n";
+		request << i2p::util::http::httpHeader(u.path_, u.host_, "1.1");
                 if (m_Etag.length () > 0) // etag
                     request << i2p::util::http::IF_NONE_MATCH << ": \"" << m_Etag << "\"\r\n";
-                if (m_LastModified.length () > 0) // if-modfief-since
+                if (m_LastModified.length () > 0) // if modified since
                     request << i2p::util::http::IF_MODIFIED_SINCE << ": " << m_LastModified << "\r\n";
                 request << "\r\n"; // end of header
                 auto stream = m_Book.getSharedLocalDestination()->CreateStream (leaseSet, u.port_);
--- a/core/AddressBook.cpp
+++ b/core/AddressBook.cpp
@@ -466,7 +466,7 @@ namespace client
         bool success = false;   
         i2p::util::http::url u (m_Link);
         i2p::data::IdentHash ident;
-        if (m_Book.GetIdentHash (u.host_, ident) && m_Book.getSharedLocalDestination())
+        if (m_Book.GetIdentHash (u.m_host, ident) && m_Book.getSharedLocalDestination())
         {
             std::condition_variable newDataReceived;
             std::mutex newDataReceivedMutex;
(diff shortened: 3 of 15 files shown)


orignal:
http reseed is not used anymore.
Instead reinventing a wheel see how it's supposed to be implemented
https://bitbucket.org/orignal/i2pd/src/c71f3ecc9019e951113ac93bcf16d73fd7400249/Reseed.cpp?at=master

But if you try to resolve #224 here, it will be crypto++/openssl messup.
Another reason of own TLS implementation was demo for my technical article describing mechanisms and implementation of SSL. But since files have been moved to another folders, all links there are dead now.

EinMByte:
I plan to use the stuff from HTTP.h/cpp to build the actual request. With boost.asio we can just indicate that we want to use TLS for any TCP socket, somewhat similar to how @orignal did it in his fork - although it will be done through a separate object or function and we should actually verify (rather than passing `boost::asio::ssl::context::verify_none`) .

This seems to be a step in the right direction, but it's only temporary (so I'll look into this and see if it's a good idea to merge it).

orignal:
You don't need to verify HTTPS certificate because .SU3 file uses own signature.
However you can't avoid crypto++/openssl mess up.

EinMByte:
@orignal There is a possibility that we will completely change to openSSLin that case (see #244). It might be possible to avoid this though, but I need to look into that.

anonimal:
>http reseed is not used anymore.

```
$ ./i2pd
i2pd starting
23:51:02/info - Version 0.10.0
23:51:02/info - data directory: /home/i2pd/.i2pd
23:51:02/info - CMD parameters:
23:51:02/info - 0  ./i2pd
23:51:02/info - HTTP Server started
23:51:02/info - 0 routers loaded
23:51:02/info - 0 floodfills loaded
23:51:02/warn - Reseed certificates not loaded. "/home/i2pd/.i2pd/certificates/reseed" doesn't exist
23:51:02/info - Downloading SU3 from http://193.150.121.66/netDb/
23:51:04/warn - Certificate for meeh@mail.i2p not loaded
23:51:04/debug - Processing file routerInfo-ko9mv5mdWrpVulCX09gDym3bx6mv4Ea3x2osGSfWZTM=.dat 713 bytes
23:51:04/info - New RouterInfo added
...
```
```
# tcpdump -lnA 'port 80'
23:51:03.190550 IP 192.168.0.2.37336 > 193.150.121.66.80: Flags [P.], seq 1:116, ack 1, win 229, length 115: HTTP: GET /netDb/i2pseeds.su3 HTTP/1.1
E.....@.@.V.......yB...P."..8..vP.......GET /netDb/i2pseeds.su3 HTTP/1.1
Host: 193.150.121.66
Accept: */*
User-Agent: Wget/1.11.4
Connection: close
23:51:03.191175 IP 193.150.121.66.80 > 192.168.0.2.37336: Flags [.], ack 116, win 457, length 0
E..(.N@.@.P...yB.....P..8..v."..P....G........
23:51:04.437045 IP 193.150.121.66.80 > 192.168.0.2.37336: Flags [P.], seq 1:499, ack 116, win 457, length 498: HTTP: HTTP/1.1 200 OK
E....O@.@.O...yB.....P..8..v."..P...FP..HTTP/1.1 200 OK
Date: Thu, 15 Oct 2015 23:51:04 GMT
Server: Apache/2.2.22 (Ubuntu)
Last-Modified: Thu, 15 Oct 2015 22:00:32 GMT
ETag: "187d75-ead2-5222bced06db4"
Accept-Ranges: bytes
Content-Length: 60114
Connection: close
I2Psu3..................................3...............meeh@mail.i2pPK..........PG.A$.........;...routerInfo-ko9mv5mdWrpVulCX09gDym3bx6mv4Ea3x2osGSfWZTM=.datZmg8c.....8.iv..p.ac.@h..F..l.B|Z.u...bGe+pJ.:Tk..D...>]).n..3..3..;..[...x.#W.
......r<..B.f=.......
```
@orignal Your confusing your fork with ours. As per a semi-recent upstream decision, yes: HTTPS is exclusively used - but we're not there yet.

>Instead reinventing a wheel see how it's supposed to be implemented

To date, I've found a fair amount of your code and approach to be sloppy and/or silly so, *no thank you*. Since you've voluntarily removed yourself from this fork, I think the rest of us can put our heads together to come up with a better implementation.

>This seems to be a step in the right direction, but it's only temporary (so I'll look into this and see if it's a good idea to merge it).

@EinMByte I agree and the more I get comfortable with the project the more I see room for much rewrite and improvement. Since this PR is not a feature or a fix, the only benefits to merging that I can see are some code re-usability and miniscule design-logic improvement that would, at the very least, help me stay on-track while we move onto better things (and it would be a small pat on my head too) :smile:

orignal:
@anonimal , not so fast, please.
Your assumption is based on wrong statements. No, I haven't removed myself from this fork.
If you want to make better implementation, why not create own fork or, even better, own project with own name.

EinMByte:
@orignal With all due respect, the best strategy here is to cooperate. Developers can disagree about design decision, but usually a compromise can be made. Anyway, let's stick to the topic. I'll try to discuss this with @anonimal 's on IRC sometime soon.

orignal:
I think the best strategy for him is to shut up about things, he doesn't know about.
A serious conversation about the situation will be done with @meeh420 soon.

anonimal:
These were collecting dust since the original PR. Because of the circus of #279, I've avoided pushing any real code until our meeting provided a viable solution for project sustainability: and it has. These commits needed to get out of my way before moving on though, so do with them what you will.

https://github.com/anonimal/i2pd/commit/66db088761a27dd9a1c930b2c46584da204b08fe is a point made at the quality that this project has allowed. I'll let it speak for itself. Yes, http serves a small purpose in i2pd but that doesn't mean it should be stolen or look disgusting. More importantly, a re-thinking of the uses of http could provide leeway for a more concrete http implementation.

I'm sure orignal will respond with his usual responses, so I'll just let that happen and wish this project the best.

orignal:
@anonimal before make such judgements about low value of http service, I would look at the history and see how much work has been done by @klondi to make these services working well.
Yes, copy&paste from anoncoin in util.cpp has to be rewritten. That's not just http, but also program options. I personally don't like that singleton there and it should be moved to RouterContext later on.

klondi:
Hey @orignal @anonimal 

I still believe HTTP Proxying is a quite critical entrypoint for i2pd users as it's most likely the first thing they'll use (although things like SOCKS5 can partially cover for that).

I started work on kproxy (https://github.com/klondi/kproxy) with the objective of being able to move that out of the equation though (mostly for another technical problem involving transparent proxying though). SOCKS5 is reasonably easy to implement as a way to proxy TCP like stuff out and my middle term objective was writting up a propper HTTP to SOCKS5 proxy. Not there yet though :(

Also maybe solutions like privoxy can fill in the gap although we should package them in an easy to use way.

PR #294, opened 2015-11-11
Updated CMake stuff to build OpenSSL version with MSVC++

mlt:
Tested with Boost 1.58 and OpenSSL 1.0.1p. You may want to use [Chocolatey](chocolatey.org/) on Windows to `cinst git kdiff3 cmake strawberryperl nasm visualstudio2015community` or whatever is missing. Then from the command line of MSVC
```
perl Configure VC-WIN32 --prefix=C:\OpenSSL-Win32
ms\do_nasm
nmake -f ms\ntdll.mak
nmake -f ms\ntdll.mak install
```
CMake will look for OpenSSL in `C:\OpenSSL-Win32` among other places.

### DIFF
--- a/Daemon.cpp
+++ b/Daemon.cpp
@@ -1,6 +1,5 @@
 #include <thread>
 #include <memory>
-#include <openssl/ssl.h>
 
 #include "Daemon.h"
 
@@ -20,6 +19,8 @@
 #include "HTTPServer.h"
 #include "I2PControl.h"
 #include "ClientContext.h"
+// ssl.h somehow pulls Windows.h stuff that has to go after asio
+#include <openssl/ssl.h>
 
 #ifdef USE_UPNP
 #include "UPnP.h"
--- a/Daemon.cpp
+++ b/Daemon.cpp
@@ -39,7 +39,7 @@ namespace i2p
 			std::unique_ptr<i2p::client::I2PControlService> m_I2PControlService;
 
 #ifdef USE_UPNP
-			UPnP m_UPnP;
+			i2p::transport::UPnP m_UPnP;
 #endif	
 		};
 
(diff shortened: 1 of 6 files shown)


mlt:
Ooops...I need to rethink how to deal with zlib.

orignal:
the lib name is libz

Also is it possible to change CMake build to produce libi2pd.a, libi2pdclient.a and link i2pd binary with them?

mlt:
It is "common" to skip *lib* prefix on libraries for MSVC builds. While you might get by reusing mingw binaries written in C, you'd be in trouble with mangled C++ names.
I believe libz is from mingw. When I build it, I get zlibd.lib and zlibd.dll . While it is possible to rename things, it will be confusing for end users why libi2pd.a built with MSVC++ won't link when using mingw.

mlt:
For now, while I'm trying to think how to wrap it all around, just comment out second find_package and remove zlib includes and library from i2pd. After it downloads and builds zlib, just `git checkout build\CMakeLists.txt` to get things back.

orignal:
I thought zlib is presented almost everywhere

mlt:
Why would you need it, if you have [this](https://msdn.microsoft.com/en-us/library/windows/desktop/hh437596%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) and [that](https://msdn.microsoft.com/en-us/library/windows/desktop/aa380255%28v=vs.85%29.aspx) or [that](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376210%28v=vs.85%29.aspx) and more maintained managed API? :eyes:

Though good point! I'll assume that first find_package will succeed everywhere but on Windows and I already know naming here.

mlt:
Also what is the point of splitting libraries? Last time I tried (long time ago) there are lots of interlinked things rendering shared library generation impossible as PE can't have unresolved symbols unlike ELF.

orignal:
Look at filelist.mk how they are split between libraries now.

If you wish to use embedded Windows crypto go ahead, sure under #ifdef WIN32

mlt:
> Look at filelist.mk how they are split between libraries now.

I saw that, but if not shared, they are all going to be linked together anyway. Anyhow.. it is on the way.
> If you wish to use embedded Windows crypto go ahead

nah... I'm good :sunglasses:

orignal:
libi2pd can be linked independently from others, libi2pdclients depends on libi2pd and i2pd depends on both

PR #295, opened 2015-11-12
Bring CMake stuff in agreement with #294 discussion

mlt:
Do you feel like decorating your code with I2PD_API stuff to make shared libi2pd possible on Windows? MSVC unlike gcc does not export all symbols.
I didn't try other platforms. Things might get broken a bit.

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -14,8 +14,7 @@ option(WITH_PCH "Use precompiled header" OFF)
 set ( CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules" )
 set ( CMAKE_SOURCE_DIR ".." )
 
-set (COMMON_SRC
-  "${CMAKE_SOURCE_DIR}/AddressBook.cpp"	
+set (LIBI2PD_SRC
   "${CMAKE_SOURCE_DIR}/Crypto.cpp"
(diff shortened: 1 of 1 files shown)


mlt:
While decorating all calls might be a PITA, *gendef*, e.g. from GTK-- (PostgreSQL version didn't work for me though [it should support DATA](http://osdir.com/ml/postgresql-pgsql-committers/2014-02/msg00144.html) ), can generate [Module Definition File](https://msdn.microsoft.com/en-us/library/28d6s79h.aspx?f=255&MSPPError=-2147217396) in place of numerous __declspec(dllexport) decorations. However you do have some variables to be exported across DLL boundary. Somehow it didn't work solely with DEF file so I had to replace `extern ...` with `extern I2PD_API ...` and alike, where I2PD_API is either __declspec(dllexport) or __declspec(dllimport). Though it is invasive, but not as invasive as messing with every single class & function to be exported. One catch is to call gendef to regenerate libi2pd.def file when things are changing. And another question is where to define macros I2PD_API . Either in a separate header and include it where necessary, or to rely on -D flags via CMake or alike.
Anyhow, are you interested in libi2pd.dll with MSVC++?

orignal:
Right now , I'm interested to have it with mingw build to be able to build it against coins applications.
MSVC++ can be added later.

PR #296, opened 2015-11-20
Make tunnels.cfg configurable

edwtjo:
This is the same change that was introduced in https://github.com/PurpleI2P/i2pd/commit/3d489a29962aef779af608b6cb311ec23552ed89 but later lost when the move back from bitbucket happened.

### DIFF
--- a/ClientContext.cpp
+++ b/ClientContext.cpp
@@ -243,13 +243,14 @@ namespace client
 	void ClientContext::ReadTunnels ()
 	{
 		boost::property_tree::ptree pt;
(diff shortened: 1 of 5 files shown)


orignal:
please be aware, that tunnel code should not appear in basic libi2pd code, it should be moved to a file libi2pdclient. I mean GetTunnelsConfigFile() should be moved from until.cpp to ClientContext similar way it's done for other client services

edwtjo:
Aha, thanks for the heads up!

orignal:
I have written the coding guideline for clarification.
https://github.com/PurpleI2P/i2pd/wiki/Coding-guideline

PR #340, opened 2016-01-23
Better checking if boost::asio::buffer works with std::array

al42and:
Currently, checking whether boost::asio::buffer constructor accepts std::array is performed by checking boost version. This fails with clang 3.4 and boost 1.54. Luckily, boost has `BOOST_ASIO_HAS_STD_ARRAY` flag specifically for this.


orignal:
good point

PR #360, opened 2016-02-04
Do not force build type

AMDmi3:
Some environments prefer to not set build type to not introduce unneeded flags, and this should be allowed.


orignal:
What's going to happen if it's not specified?

AMDmi3:
It just won't append CMAKE_CXX_FLAGS_DEBUG and all other kinds of flags. Default flags like CMAKE_CXX_FLAGS still work.

PR #397, opened 2016-02-28
Irc ip to b32

xcps:


majestrate:
does the irc tunnel only do the filter 1 time or is it done for each line?

orignal:
@majestrate that's a good question probably we should do it only once and then pass "as is"

PR #400, opened 2016-03-02
Add Transit/Client Tunnel Count to web ui

majestrate:


orignal:
size_t CountxxxxTunnels() should be 'const'

PR #436, opened 2016-03-24
add option for toggling ipv4

majestrate:


orignal:
You also have to care about resolvers in Transport.cpp to make sure that appropriate address is selected.

majestrate:
noted, wouldn't binding to just ipv6 work if ipv4 is disabled?

orignal:
It might get resolved to ipv4 and try to connect to it despite it's disabled. 
Opposite situation to what dadada seen with ipv6.

PR #440, opened 2016-03-26
add syslog option for logging

### LINKED ISSUE #423: Add possibility to log to syslog
It will be useful if i2pd could log to syslog instead of file and stout only.

majestrate:
attempts to fix #423


majestrate:
aha, this would break win32, let me update this a bit

PR #443, opened 2016-03-27
streaming fixes

majestrate:
not sure if these were errors or not but so far changes make my irc connection much more stable.


orignal:
Packet * p = new Packet;
where is it going to be deleted?
You must delete it after SendPackets

PR #450, opened 2016-03-28
i2p family cert file

manasb:
Hello,

This is my I2P router family's certificate file.

Thanks


orignal:
what's your family name? just i2p?

PR #481, opened 2016-04-26
Webiface

weekendi2p:
show family name in web interface, under the status line


weekendi2p:
no need to invade the SetFamily() function
i'll come back later with a more elegant way.

PR #526, opened 2016-06-19
Android setForeground fix + notific.icon + fs crash fix + backg.thread restored

hypnosis-i2p:


hypnosis-i2p:
This https://github.com/PurpleI2P/i2pd/issues/518 can be closed.

PR #580, opened 2016-07-24
Reloadconfig on SIGHUP

mikalv:
Like Tor and other daemons, i2pd should be able to reload config without a fully restart. This is usually done with a SIGHUP signal. Implemented and tested on OSX.


mikalv:
@orignal OK for merging?

mikalv:
Ugh, NVM. Based on wrong branch. I'll create a new one.

PR #590, opened 2016-08-04
tiny commit to check pulling

brain5lug:
Would you kindly accept my tiny pulling request? I want to check if you can pull my changes.


majestrate:
good catch, let's let the CI finish

PR #615, opened 2016-08-19
Find miniupnpc library as well as header

AMDmi3:


AMDmi3:
Just noticed that this has broken appveyor build. It should probably be provided with path to third party library directory.

orignal:
appveyor is always broken

AMDmi3:
It now complains to UPNP_LIBRARY, so this change had broken it further :)

PR #660, opened 2016-10-06
Unused parameter warnings removal

brain5lug:


orignal:
Leave it as is for now.
They will be used later

PR #662, opened 2016-10-10
Merge recent udp tunnel changes

majestrate:
fixes memory leak


majestrate:
okay now should be ready for merge

PR #666, opened 2016-10-11
Add addressbook.subscriptions option

l-n-s:
Add addressbook.subscriptions option + delete default subscription constant + improved docs


l-n-s:
Fail

PR #676, opened 2016-10-20
added spec and service files

BOPOHA:
There are all you need to build RPMs in COPR repo.
For example https://copr.fedorainfracloud.org/coprs/vorona/i2pd/

### DIFF
--- a/i2pd.service
+++ b/i2pd.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=I2P router
+After=network.target
+
+[Service]
+User=i2pd
+Group=i2pd
+Type=simple
+ExecStart=/usr/bin/i2pd --service
+PIDFile=/var/lib/i2pd/i2pd.pid
+Restart=always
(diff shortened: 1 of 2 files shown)


l-n-s:
So, it is like PPA in Ubuntu, but for Fedora? Are those files required to be in our root directory, or may be we can find some better directory for it?

BOPOHA:
> So, it is like PPA in Ubuntu, but for Fedora?

Yes, it works for all redhat-based systems.
( as another way, you can build Ubuntu, Debian, Centos, etc-etc-etc, in one place, here https://build.opensuse.org/)

> or may be we can find some better directory for it?

well, If it is really necessary, i think we can. Just move it and l'll fix paths.

l-n-s:
Cool. I guess `contrib/rpm` directory is the best place for such files.

orignal:
@l-n-s yes, please move it there

l-n-s:
done

PR #681, opened 2016-10-20
remove to call C functions and use C++ std::fstream

sizeofvoid:
- Point std{in,out,err} descriptors to /dev/null (C++ wey)


orignal:
I'm not sure we can redefine descriptors 0,1,2 this way.

orignal:
It doesn't do what it's supposed to. We must make sure that these file descriptors are redirected properly rather than streams to make sure they are not taken for sockets, etc.

sizeofvoid:
i2pd-2.10.0/DaemonLinux.cpp:79:24: error: expression is not assignable
                stdin  = freopen("/dev/null", "r", stdin);
                ~~~~~  ^
i2pd-2.10.0/DaemonLinux.cpp:80:12: error: expression is not assignable
                                stdout = freopen("/dev/null", "w", stdout);
                                ~~~~~~ ^
i2pd-2.10.0/DaemonLinux.cpp:81:12: error: expression is not assignable
                                stderr = freopen("/dev/null", "w", stderr);

clang -v
clang version 3.9.0 (tags/RELEASE_390/final)
Target: amd64-unknown-openbsd6.0
Thread model: posix

orignal:
Target: amd64-unknown-**openbsd**6.0
Add #if  !defined(OPENSBD) or something like this

sizeofvoid:
Is this ´#define´ to support OpenBSD as target system acceptable as pull request?

orignal:
Sure it is

PR #695, opened 2016-11-01
add optional websocket ui

majestrate:
optional because in build it pulls in another library for websocket handling
I can refactor it later to not pull that websocket library in or perhaps bundle it as it's just a header library.

### DIFF
--- a/ClientContext.cpp
+++ b/ClientContext.cpp
@@ -290,7 +290,7 @@ namespace client
 		}
 		return infos;
 	}
-
+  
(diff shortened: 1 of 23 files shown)


l-n-s:
Just wondering, can we websocket-enable i2pcontrol API and manage i2pd from regular HTTPService?

l-n-s:
and the same with SAM and other APIs, to build I2P websocket apps

majestrate:
you could but that would need to be implemented

PR #697, opened 2016-11-03
systemd unit, debian patch fix, makefile last-dist added, package log&pid folder edited

atnaguzin:


atnaguzin:
I will rework some things, and then rePR again.

PR #759, opened 2017-01-01
Build DEB package with CMake and uPNP

v2e:
- Build Debian package using CMake instead of the shipped Makefile
- enable uPNP support during build

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: i2pd
 Section: net
 Priority: extra
 Maintainer: hagen <hagen@i2pmail.org>
-Build-Depends: debhelper (>= 9.0.0), dpkg-dev (>= 1.16.1~),
+Build-Depends: cmake (>= 2.8.12), debhelper (>= 9.0.0), dpkg-dev (>= 1.16.1~),
  gcc (>= 4.7) | clang (>= 3.3),
  libboost-system-dev (>= 1.46),
  libboost-date-time-dev,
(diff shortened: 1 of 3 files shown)


orignal:
We don't enable UPnP intentionally

v2e:
> We don't enable UPnP intentionally
Why? I tested on a couple of my PCs, and with UPNP it works much more effective (more transit traffic).

And what about the CMake build itself? It is independent of UPNP.

r4sas:
About cmake:
If you will give to me 10 reasons to build deb packages with cmake, I'll think about this

orignal:
UPnP requires extra dependancy, and most of Linux users don't want it, because they know how to forward ports manually

majestrate:
from what I understand the cmake build files aren't really used much

orignal:
FreeBSD build uses it

v2e:
@r4sas This repository contains the CMake files already. So I just suggested to use them during the DEB package build. It was nothing more but a suggestion. So it is perfectly OK to decline it. And sure enough I am not going to "give to you 10 reasons".

@orignal You may be right about the linux users knowing how to forward ports by hand. Still, I was considering the following:

1. Most torrent-clients (as well as some other software) use UPNP anyway, so the library may be there on the user's PC anyway. And this library is a standard one from the Debian repository.
2. The I2P documentation seems to recommend the user NOT to use the same port all the time. If I am not mistaken, it is recommended to use a random port. In such case it would be not too convenient to forward a new port each time by hand.

This being said, I'd like to emphasize that by no means I insist on these commits. I will use them on my own system, but it is totally up to you to decide what is going to enter the code tree. And thank you guys for working on I2PD! Please continue. :)

r4sas:
@v2e, =)

PR #771, opened 2017-01-07
merge recent features and bugfixes

majestrate:
includes following new features:

* optionally use loopback addresses in server tunnels
* http outproxy via local socks/http proxy specified via url, i.e. socks://127.0.0.1:9050/ or http://127.0.0.1:8118/ , .i2p tld support not yet added.
* use larger buffer in tcp/ip pipe so that upstream socks proxy goes faster
* udp tunnel fixups 
* don't spam websocket events if enabled
* allow anonymous name lookups from SAM
* use lowest SAM version
* add hook for tunnel build result in i2lua


majestrate:
review desired before merge

PR #797, opened 2017-02-08
Added systemd support for debian package

kayrus:
It is suitable for modern deb based systems like Debian Jessie (8) or Ubuntu Xenial (16.04).

Logs are written directly into systemd-journald which helps to debug i2pd: `journalctl -f -u i2pd`. Actually there is no need in logrotate rules, but I left them for backward compatibility.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -5,6 +5,7 @@ Maintainer: hagen <hagen@i2pmail.org>
 Build-Depends: debhelper (>= 9.0.0), dpkg-dev (>= 1.16.1~),
  gcc (>= 4.7) | clang (>= 3.3),
  libboost-system-dev (>= 1.46),
+ dh-systemd (>= 1.5),
  libboost-date-time-dev,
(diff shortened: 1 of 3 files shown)


r4sas:
We already have it: https://github.com/PurpleI2P/i2pd/tree/openssl/contrib/debian

kayrus:
@r4sas it is not obvious though.

majestrate:
wouldn't this need to be amended to not run i2pd as root?

kayrus:
@majestrate true. all this stuff was already implemented here: https://github.com/PurpleI2P/i2pd/blob/openssl/contrib/debian/i2pd.service

kayrus:
I've updated i2pd.service file. If you don't want to merge this PR, just close it.

majestrate:
A merge should be fine given that we have someone to maintain the debian package, that is the part that is foggy for me.

PR #900, opened 2017-07-06
Fucking monero.

StPrimateCoder:
Fucking the "Monero" organisation simply stealing your a code. 
I think it better option for problem. Because "Monero" can use other name I wrote 'or similar projects' . I did this. Now you must accept or refuse about my idea. You may think about my idea 20 years and more.

Do good.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -16,7 +16,7 @@ provided with the distribution.
 to endorse or promote products derived from this software without specific prior written
 permission.
 
-4. You do not access for use source codes if you Monero. 
+4. You do not access for use source codes if you Monero or similar project. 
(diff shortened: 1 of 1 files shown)


majestrate:
would prefer to not perpetuate e-drama.

orignal:
@majestrate although this PR definitly will not be accepted, however he has a point. We might formulate the same more politely, like our code can't be used in any project under Monero's license

majestrate:
You could always do GPL-3 but that would put i2pd at a massive disadvantage.

v2e:
@majestrate What kind of disadvantage?

PR #917, opened 2017-08-09
Don't do peertest if nat=false (#915)

l-n-s:
**REVIEW PLEASE**

Added ``Transports::m_IsNAT`` field and it's getter and setter, so we don't have to read config all the time.

### DIFF
--- a/libi2pd/Transports.cpp
+++ b/libi2pd/Transports.cpp
@@ -110,7 +110,7 @@ namespace transport
 	Transports transports;
 
 	Transports::Transports ():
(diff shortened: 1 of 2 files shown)


majestrate:
question: is `Transports::SetNAT` used anywhere outside that class?

l-n-s:
@majestrate I don't think so. So you mean it is not really required, and we can set the field directly?

majestrate:
@l-n-s that was the question i was getting at, is it really used elsewhere?

majestrate:
it's a great PR just a nitpick

l-n-s:
Good nitpick! Looks like getter and setter are useless in this situation, so I'll fix that PR asap.

PR #933, opened 2017-08-27
Fix deformed json result with an extra comma

myfingerhurt:
The deformed data in I2PControl :
`{"id":1,"result":{"i2p.router.net.bw.in":48,"i2p.router.net.bw.out":48,},"jsonrpc":"2.0"}`
You can see there is an extra comma behind the number 48.


majestrate:
neato

Issue #957, opened 2017-09-10
[macOS] Missing dependencies in release package.


p-groarke:
Could you bundle the dependencies in your release builds, and set appropriate dylib search paths? Or better yet, link statically with non system libraries.

Notably:

```
./i2pd
dyld: Library not loaded: /usr/local/lib/libcrypto.1.1.dylib
  Referenced from: /Volumes/User/me/Downloads/i2pd_2.15.0_osx/./i2pd
  Reason: image not found
Abort trap: 6
```

List of non system deps:

```
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
	/usr/local/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/local/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	libboost_system.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_date_time.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_filesystem.dylib (compatibility version 0.0.0, current version 0.0.0)
	libboost_program_options.dylib (compatibility version 0.0.0, current version 0.0.0)
```

It is strange that you ship with readily available system dependencies (libc++, libsystem.B), which are both guaranteed to be on a mac system. But omit additional dependencies.

Thanks

orignal:
Thank you for pointing me. Will add

majestrate:
i think i should invest some time into adding mac os application bundles to makefiles.

p-groarke:
My advice is make everything static (boost + openssl) and leave out the system libs. You aren't actually using the shipped libc++ or System.B anyways (the executable uses the ones in /usr/lib).

That should make the release much less prone to breaking and less time consuming. Which is nice since as you mentioned, macOS isn't popular.

majestrate:
I fixed the homebrew makefile such that it can do static linking: https://github.com/majestrate/i2pd/tree/macos-static

    i2pd:
        /usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.5)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)

orignal:
@p-groarke  I'm not good in Macs. I make these build using hackintosh in a virtual machine.

[...]

PR #958, opened 2017-09-11
fix up homebrew makefile to allow static linking

majestrate:
addresses #957


majestrate:
I also see some problems with the qt build on mac os so i'll fix those up too

PR #960, opened 2017-09-18
Small patches

l-n-s:
fixed typo + added full date to logs


orignal:
we don't need date for every single line

PR #961, opened 2017-09-19
fixed typo + added optional full date in logs `logclftime=true`

l-n-s:


orignal:
Please don't read config in Log.cpp but set it as options like i2p::log::Logger().SendTo (logStream) in api.cpp

PR #965, opened 2017-09-27
try fixing leak

majestrate:
ed25519 code has a subtle leak.

`BN_ctx_start` is called more times than `BN_ctx_end`


libi2pd/Signature.cpp
```diff
@@ -316,7 +316,6 @@ namespace crypto
 
 			bool IsOnCurve (const EDDSAPoint& p, BN_CTX * ctx) const
 			{
-				BN_CTX_start (ctx);
```
majestrate: This right here is the fix

PR #969, opened 2017-09-30
atomics have been used for threads stop

brain5lug:
boolean used to stop treads have been replaced with atomics. Those flag without event volatile could be optimized by compiler with compiler unpredictable way. Check could be completely omitted as simple boolean are treated as single thread used.


orignal:
Please don't produce such overhead. It's atomic already

PR #988, opened 2017-11-04
Follow XDG Base Directory Specification for datadir (FIX #971)

### LINKED ISSUE #971: Follow XDG standard for creating datadir in $HOME
https://standards.freedesktop.org/basedir-spec/latest/ar01s03.html

https://wiki.archlinux.org/index.php/XDG_Base_Directory_support

i2pd should search `$HOME/.config/i2pd` for config files and `$HOME/.local/share/i2pd` for data files.

(it is similar standard as Windows has with %APPDATA%)

l-n-s:


orignal:
where will transient files like netdb go?

l-n-s:
kk, hold this PR, will update later

majestrate:
Is there any particular reason we need to conform with XDG stuff? right now the ~/.i2pd setup works fine

splitting it up sounds like a major asspain.

l-n-s:
Looks too hard for me atm, maybe will try later

@majestrate it looks like a good idea to split config directory from datadir, almost all programs do that. For example, one can store data in tmpfs to reduce disk usage on a raspberry pi - and still keep all configuration/keys.

PR #1027, opened 2017-12-09
Use transient destinations

majestrate:
use transient destinations by default instead of shared local destination when client tunnel keys not specified.

someone on my irc thought this was the default but it wasn't.


orignal:
we should also introduce "shared"

PR #1030, opened 2017-12-11
Reorder preprocessor conditions in libi2pd/I2PEndian.h

AMDmi3:
The problem is that __FreeBSD_kernel__ may be defined on FreeBSD as
well, while it always needs <sys/endian.h>

In fact, you shouldn't check for __FreeBSD_kernel__ at all, as the code has nothing to do with whether the code runs under FreeBSD kernel or not.

### DIFF
--- a/libi2pd/I2PEndian.h
+++ b/libi2pd/I2PEndian.h
@@ -3,10 +3,10 @@
 #include <inttypes.h>
 #include <string.h>
 
-#if defined(__linux__) || defined(__FreeBSD_kernel__) || defined(__OpenBSD__)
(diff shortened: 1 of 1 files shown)


AMDmi3:
Thanks! Please note the suggestion in PR description.

orignal:
FreeBSD_kernel check was not implemented bt me, but kytv I believe. That's about debian with freebsd's kernel

AMDmi3:
+@kytvi2p

orignal:
He has left I2P almost 2 years ago

PR #1031, opened 2017-12-11
Don't use deprecated boost::asio::ssl::context ctor

AMDmi3:
This was removed in boost 1.66, in prior versions the service
argument was not used


r4sas:
Dont forget that older distibutions still uses boost 1.56, for example Debian 8.
Please rewrite with `#ifdef` definitions only fot boost >1.66.0

orignal:
You must check for boost version using BOOST_VERSION define. Please rewrite it this way. Look at I2PControl for example and fix.

AMDmi3:
That is not needed. The 2 argument ctor was deprecated (and 1 argument ctor available) for ages, since at least boost 1.47.

PR #1036, opened 2017-12-16
Fix typo (confrimed -> confirmed)

KenanSulayman:
I couldn't stop my OCD from being triggered when I saw that in the terminal, so here is a quick PR for the typo.

Thanks for the project.

🍺


orignal:
Please fix in openssl as well

PR #1060, opened 2018-01-15
Inherit _publicly_ from shared_from_this in I2PService

inetic:
We're using I2PServerTunnel and I2PClientTunnel (which both inherit from I2PService) in our project. We also have our custom "sockets" which are basically pairs:

```
struct my_socket {
    asio::ip::tcp::socket socket;
    shared_ptr<I2PService> i2p_tunnel;
};
```
The reasoning being that while an instance of `my_socket` is present, the tunnel won't get closed.

If the `enable_shared_from_this` base is private, we can't create `shared_ptr<I2PService>` and we need ugly workarounds such as:

```
struct my_socket {
    asio::ip::tcp::socket socket;
    shared_ptr<unique_ptr<I2PService>> i2p_tunnel;
};
```

### DIFF
--- a/libi2pd_client/I2PService.h
+++ b/libi2pd_client/I2PService.h
@@ -14,7 +14,7 @@ namespace i2p
 namespace client
 {
 	class I2PServiceHandler;
-	class I2PService : std::enable_shared_from_this<I2PService>
+	class I2PService : public std::enable_shared_from_this<I2PService>
 	{
 		public:
 			typedef std::function<void(const boost::system::error_code &)> ReadyCallback;


majestrate:
Looks good will merge after CI check pass.

My I ask which project is using this? (curious)

inetic:
Thanks, it's for a project called [ouinet](https://github.com/equalitie) (the name of the project and its submodules are still under consideration, + we're very bad at naming things :) ). Its a project that tries to utilize different distributed network projects (currently IPFS, GNUnet and I2Pd) to help serve http content when direct ways are not available.

PR #1067, opened 2018-01-22
Update rpm spec and systemd unit

l-n-s:
Added default configs and reseed certificates to the package + updated systemd unit

@BOPOHA can you give a review?


l-n-s:
I also wonder, why do we need a separate -systemd package? i2pd is a system daemon, not a userspace tool.

PR #1132, opened 2018-02-27
Make links relative

micdm:
I'm using Nginx to proxying all local web-GUI into single website. So i.e. when Service1 is listening on localhost:7000 then it is available by http://servername/service1, Service2 from localhost:8000  going to http://servername/service2 and so on.
The same works for i2pd web interface too excepting menu links that directing to site root instead of current directory. For example: i2pd proxied address is http://servername/i2pd, and links should lead to http://servername/i2pd?page=blabla but now they lead to http://servername/?page=blabla.
So I'm just proposing to remove first slash from all links to make them relative.

### DIFF
--- a/daemon/HTTPServer.cpp
+++ b/daemon/HTTPServer.cpp
@@ -170,16 +170,16 @@ namespace http {
 			"<div class=header><b>i2pd</b> webconsole</div>\r\n"
 			"<div class=wrapper>\r\n"
 			"<div class=left>\r\n"
-			"  <a href=\"/\">Main page</a><br>\r\n<br>\r\n"
-			"  <a href=\"/?page=" << HTTP_PAGE_COMMANDS << "\">Router commands</a><br>\r\n"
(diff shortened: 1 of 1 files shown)


r4sas:
I think you must fix links by yourself in nginx with ngx_http_subs_filter_module, so I'm against apllying that changes.

micdm:
> I'm against apllying that changes.
Is there some ideological reason? Actually the current behavior will not even be changed.
I will definitely not use Substitutions module as it's non-default one. Maybe some rewrite rules will help...
Anyway thanks for attention, you're doing great!

majestrate:
using nginx to rewrite html is sucky and this patch looks fine.

orignal:
web console is not supposed to be used in projects like this. It's diagnostic tool only.
I2PControl must be used instead

micdm:
OK I see your point. Probably a better approach could be adding of new parameter to "http" config section like "linkprefix". So anybody can change it for his needs without recompiling.
I'm not sure if I can do that by myself :smiley: but if it would be suitable I'll try.

PR #1133, opened 2018-02-28
Time experemental options

borned-mind:
not tested


libi2pd/Timestamp.h
```diff
@@ -3,30 +3,44 @@
...
+
+			auto tmp_time = std::chrono::duration_cast<std::chrono::seconds>(
 				 std::chrono::system_clock::now().time_since_epoch()).count ();
+
+			return (Time_Correcting || Time_UseNTP)  ? tmp_time + i2p::util::g_TimeOffset  : tmp_time;
+		}
+
+		inline uint64_t GetMillisecondsSinceEpoch ()
+		{
+			return GetSecondsSinceEpoch()*1000;
```
majestrate: we need millisecond granularity in some parts of the code.

orignal:
GetMillisecondsSinceEpoch () is used in many other places. This change will break many things

PR #1135, opened 2018-03-03
I2PControl: add new method ClientServicesInfo

l-n-s:


daemon/I2PControl.cpp
```diff
@@ -92,6 +93,14 @@ namespace client
...
 		m_NetworkSettingHandlers["i2p.router.net.bw.in"]  = &I2PControlService::InboundBandwidthLimit;
 		m_NetworkSettingHandlers["i2p.router.net.bw.out"] = &I2PControlService::OutboundBandwidthLimit;
+
+		// ClientServicesInfo
+		m_ClientServicesInfoHandlers["I2PTunnel"] = &I2PControlService::I2PTunnelInfoHandler;
+		m_ClientServicesInfoHandlers["HTTPProxy"] = &I2PControlService::HTTPProxyInfoHandler;
+		m_ClientServicesInfoHandlers["SOCKS"] = &I2PControlService::SOCKSInfoHandler;
+		m_ClientServicesInfoHandlers["SAM"] = &I2PControlService::SAMInfoHandler;
+		m_ClientServicesInfoHandlers["BOB"] = &I2PControlService::BOBInfoHandler;
+		m_ClientServicesInfoHandlers["I2CP"] = &I2PControlService::I2CPInfoHandler;
```
majestrate: make all these case insensitive IMO

daemon/I2PControl.h
```diff
@@ -98,6 +100,15 @@ namespace client
 			void InboundBandwidthLimit  (const std::string& value, std::ostringstream& results);
 			void OutboundBandwidthLimit (const std::string& value, std::ostringstream& results);
 
+			// ClientServicesInfo
+			typedef void (I2PControlService::*ClientServicesInfoRequestHandler)(std::ostringstream& results);
```
majestrate: make this use std::function maybe?

majestrate: nvm it doesn't need that

PR #1137, opened 2018-03-04
Makefile: add detection for GNU userspace

yangfl:


orignal:
bsd builds are supposed to use clang rather rather than gcc

orignal:
Never mind

PR #1142, opened 2018-03-05
Create LogsDirectory in i2pd.service

ffontaine:
Create /var/log/i2pd through LogsDirectory parameter of systemd and set
its permission to 0700 through LogsDirectoryMode. Indeed, this directory
must be created with the correct permission as it is used in ExecStart
command

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/contrib/i2pd.service
+++ b/contrib/i2pd.service
@@ -8,6 +8,8 @@ User=i2pd
 Group=i2pd
 RuntimeDirectory=i2pd
 RuntimeDirectoryMode=0700
+LogsDirectory=i2pd
+LogsDirectoryMode=0700
 Type=simple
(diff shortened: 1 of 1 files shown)


l-n-s:
Ain't this supposed to be done by OS specific package?

ffontaine:
You're right, I'm adding i2pd to buildroot. At the start, I wanted to reuse your i2pd.service file but finally I copied it to manage my own one due to some OS specific parts such as i2pd being installed into /usr/bin instead of /usr/sbin, RuntimeDirectory which created i2pd in /run and not /var/run and finally this latest problem about /var/log/i2pd. I thought that this latest update could be applied upstream but if you prefer to refuse this PR, it's also ok.

PR #1160, opened 2018-04-21
A small pach into main trunk to build cponsole Android i2pd binary

unlnown542a:
i created android/i2pd folder with slightly ammended Android.mk and Application.mk
Basically Android i2pd command line binary builds find with UnixDameon.ccp with smal ifdef for locking.. since in Dameon.h for Andoird is a kind of dummy placeholder. i removed so it default to linux class and all works - tested by quite good amoiunt of time. So you decid or you have it in the source tree or the guys lke me loving CLI tools will appply own pathches to ahve CLI andropid binary which is really coool as linux one.


unlnown542a:
to build i2pd binary for Android just in android direcgtory laucnh ndk-build -I i2pd which makes ndk to look into folder i2pd for mk files.

PR #1170, opened 2018-04-26
Separated buld directory for Android i2pd only binary

unlnown542a:
So now i do not touch gradle build - its ok.
added define ANDROID_BINARY to include DaemonLinux and all menus in Webconsole


unlnown542a:
to build it - change dir to android_binary_only
edit jni/Application.mk - define path to static libs (same as for normal ANdroid buld)
run ndk-build
find i2pd binary in libs/.../i2pd

PR #1190, opened 2018-05-28
fixes #1124 and calls the ready callbacks if the tunnel gets ready an…

### LINKED ISSUE #1124: I2PService::HandleReadyCheckTimer times out timers that are not timed out
Here https://github.com/PurpleI2P/i2pd/blob/openssl/libi2pd_client/I2PService.cpp#L90

It says if(itr->second >= now) then send a timeout error to the handler. But Doesn't itr->second >= now means that the handler time out time is still in future? shouldn't it be if (itr->second <= now)? Am I missing some logic here?

vmon:
…d no timeout is set.

### DIFF
--- a/libi2pd_client/I2PService.cpp
+++ b/libi2pd_client/I2PService.cpp
@@ -14,6 +14,7 @@ namespace client
 		m_LocalDestination (localDestination ? localDestination :
 			i2p::client::context.CreateNewLocalDestination (false, I2P_SERVICE_DEFAULT_KEY_TYPE)),
 			m_ReadyTimer(m_LocalDestination->GetService()),
+			m_ReadyTimerTriggered(false),
 			m_ConnectTimeout(0),
 			isUpdated (true)
 	{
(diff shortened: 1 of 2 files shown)


vmon:
currently if user set a callback function to be called when the i2ptunnel is ready to accept or make connection using I2PService::AddReadyCallback but *does not* set a timeout limit for the tunnel getting ready operation using I2PService::SetConnectTimeout, the callback function is never called back even if the tunnel gets ready.

This behavior is not intuitive and is not documented anyway.

This patch offer a solution by changing the the behavior of the code from (a) to (b):

(a) If no no timeout is set for the tunnel then no ReadyCallback is called when the tunnel gets ready.

(b) If no timeout is set for the tunnel all ReadyCallbacks are called after the tunnel gets ready.

Another (better) solution would be to set the timeout for the ready callback at the same time when one sets the callback changing the prototype of I2PService::AddReadyCallback from:

    void I2PService::AddReadyCallback(ReadyCallback cb)

to 

    void I2PService::AddReadyCallback(ReadyCallback cb, uint32_t timeout)

If this is the preferred way of solving the problem I can submit another patch for this fix.

PR #1201, opened 2018-06-24
Revert "Read explicitPeer config settings into params"

orignal:
Reverts PurpleI2P/i2pd#1200


orignal:
it breaks functionality

shakamd:
Hmm, seems like a minimal change. What breaks?

PR #1226, opened 2018-08-18
makefile: linux: add -latomic

radfish:
Tested on Arch Linux (x86_64, armv7h) and Debian unstable (on PPC) with gcc 8.2.0. On Arch Linux
on x86_64 it built without this, but also builds with this. Without this
patch On Debian unstable on PPC linking fail with undefined symbols:
/usr/include/c++/8/bits/atomic_base.h:396: undefined reference to `__atomic_load_8'

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -44,7 +44,7 @@ ifeq ($(USE_STATIC),yes)
 	LDLIBS += -lpthread -static-libstdc++ -static-libgcc -lrt -ldl
 	USE_AESNI := no
 else
-	LDLIBS = -lcrypto -lssl -lz -lboost_system -lboost_date_time -lboost_filesystem -lboost_program_options -lpthread
+	LDLIBS = -lcrypto -lssl -lz -lboost_system -lboost_date_time -lboost_filesystem -lboost_program_options -lpthread -latomic
 endif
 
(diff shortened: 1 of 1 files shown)


yangfl:
But IIRC it won't build on Debian x86_64 with this.

radfish:
On Fri, Aug 17, 2018 at 08:54:01PM -0700, yangfl wrote:
> But IIRC it won't build on Debian x86_64 with this.

Ok, thanks for info. I'll make a chroot for Debian 9 and Debian unstable
and test.

orignal:
https://github.com/PurpleI2P/i2pd/commit/b486d1cd27a699fa6e358dc8e11e4bdd7f054313#diff-0300019dc8779a719c948b7dc78ec966

Because it had some issues

orignal:
compiles fine on Ubuntu 16.04
Probably I had this issue with 14.04. Let me check Monday

radfish:
On Sat, Aug 18, 2018 at 11:18:42AM -0400, redfish wrote:
> PS. I'm testing cmake build on Debian sid on PPC now to see if
> it correctly adds -latomic automatically. But would be good
> to fix make build there too.

cmake build on Debian sid on PPC does add -latomic automatically and
works. Still should fix the make build too.
>

PR #1268, opened 2018-11-09
Added configurable web console URL path

sokolas:
URL path to web console is now configurable instead of hard-coded "/" to allow better integration with reverse proxies (like nginx)


r4sas:
https://github.com/PurpleI2P/i2pd/pull/1132#issuecomment-368900248

PR #1278, opened 2019-01-06
Fix compilation without deprecated OpenSSL APIs

neheb:
More info: https://github.com/openwrt/packages/pull/7845

### DIFF
--- a/daemon/I2PControl.cpp
+++ b/daemon/I2PControl.cpp
@@ -569,8 +569,8 @@ namespace client
 			EVP_PKEY_assign_RSA (pkey, rsa);
 			X509 * x509 = X509_new ();
 			ASN1_INTEGER_set (X509_get_serialNumber (x509), 1);
-			X509_gmtime_adj (X509_get_notBefore (x509), 0);
-			X509_gmtime_adj (X509_get_notAfter (x509), I2P_CONTROL_CERTIFICATE_VALIDITY*24*60*60); // expiration
+			X509_gmtime_adj (X509_getm_notBefore (x509), 0);
+			X509_gmtime_adj (X509_getm_notAfter (x509), I2P_CONTROL_CERTIFICATE_VALIDITY*24*60*60); // expiration
 			X509_set_pubkey (x509, pkey); // public key
 			X509_NAME * name = X509_get_subject_name (x509);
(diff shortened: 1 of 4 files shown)


orignal:
what is this change for? 
you don't need to check LEGACY_OPENSSL and CHAHA20 together, because LEGACY_OPENSSL means openssl below 1.1.0

neheb:
Yes? In OpenSSL 1.1, chacha and poly1305 can be compile time disabled.

neheb:
Space. OpenWrt is an embedded distro.

orignal:
Please elaborate. Do you believe own implementation of chacha20/poly1305 produces smaller binary vs. if we call it from openssl in case of 1.1?

neheb:
1.1

no-chacha no-poly1305 can be passed to OpenSSL's Makefile to disable it at compile time. Meaning, those functions are completely unavailable in libssl (or libcrypto) and cannot be used.

This pull request is only a fix for compilation.

orignal:
They are disable automatically because LEGACY_OPENSSL is false in case of 1.1 and up, so this code never gets compiled.
https://github.com/PurpleI2P/i2pd/blob/openssl/libi2pd/Crypto.h#L24
Seems you PR does nothing but adds more code

neheb:
...................................

Do I need to send you logs or something? This is definitely failing when chacha20 or poly1305 are disabled at the openssl level. This has nothing to do with the version of openssl.

neheb:
Oh actually here: https://gist.github.com/neheb/81e1d7c55fd8d3c5556892999e393b50

orignal:
What version of openssl do you build with? 1.1.0?

neheb:
1.1.1. 1.1.0 should give identical results.

orignal:
If I understood you right, openssl 1.1 for openwrt doesn't contain AEAD/Chacha20/Poly1305.
Let's wait for @LLE8

neheb:
By default it does. However it can be disabled through "make menuconfig"

orignal:
I have added chacha20/poly1305 check. 
https://github.com/PurpleI2P/i2pd/commit/f74b27c58c9416083bcab5c7a8c0ff2f4bc57586
Please verify on openwrt, fix the rest and create new PR

PR #1281, opened 2019-01-06
Fix compilation without deprecated OpenSSL APIs

neheb:
Leftover stuff from https://github.com/PurpleI2P/i2pd/pull/1278

### DIFF
--- a/libi2pd/Crypto.cpp
+++ b/libi2pd/Crypto.cpp
@@ -1246,7 +1246,9 @@ namespace crypto
 	void InitCrypto (bool precomputation)
 	{
 		i2p::cpu::Detect ();	
+#if LEGACY_OPENSSL
 		SSL_library_init ();
+#endif
 /*		auto numLocks = CRYPTO_num_locks();
 		for (int i = 0; i < numLocks; i++)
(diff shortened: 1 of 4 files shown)


libi2pd/Crypto.cpp
```diff
@@ -1246,7 +1246,9 @@ namespace crypto
 	void InitCrypto (bool precomputation)
 	{
 		i2p::cpu::Detect ();	
+#if LEGACY_OPENSSL
```
orignal: how do we initialize in case of 1.1?

neheb: You don't. 1.1 handles this automatically. See: https://www.openssl.org/news/changelog.html#x12

r4sas: That?

 ```
 *) Added support for auto-initialisation and de-initialisation of the library.
     OpenSSL no longer requires explicit init or deinit routines to be called,
     except in certain circumstances. See the OPENSSL_init_crypto() and
     OPENSSL_init_ssl() man pages for further information.
     [Matt Caswell]
```

neheb: Yes. OPENSSL_init_ssl only needs to be called in case of non-default options such as initializing the engine.

PR #1285, opened 2019-01-20
Make persistent storage synchronization interval configurable 

hakunamtu:
The option allows users to make disk io/memory usage tradeoffs themselves. Zero interval means saving only at exit, profiles are stored in memory.

### DIFF
--- a/libi2pd/NetDb.cpp
+++ b/libi2pd/NetDb.cpp
@@ -26,7 +26,7 @@ namespace data
 {
 	NetDb netdb;
 
-	NetDb::NetDb (): m_IsRunning (false), m_Thread (nullptr), m_Reseeder (nullptr), m_Storage("netDb", "r", "routerInfo-", "dat"), m_PersistProfiles (true), m_HiddenMode(false)
+	NetDb::NetDb (): m_IsRunning (false), m_Thread (nullptr), m_Reseeder (nullptr), m_Storage("netDb", "r", "routerInfo-", "dat"), m_PersistSyncInterval(60), m_PersistProfiles (true), m_HiddenMode(false)
 	{
 	}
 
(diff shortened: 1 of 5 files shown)


orignal:
will be merged after release 2.23

libi2pd/NetDb.cpp
```diff
@@ -123,13 +126,19 @@ namespace data
...
 					lastManageRequest = ts;
 				}
-				if (ts - lastSave >= 60) // save routers, manage leasesets and validate subscriptions every minute
+				if (ts - lastLeasesetsManage >= 60) // manage leasesets and validate subscriptions every minute
 				{
-					if (lastSave)
+					if (lastLeasesetsManage)
 					{
-						SaveUpdated ();
+						RemoveExpired ();
```
r4sas: Removing every minute?

hakunamtu: As the old `SaveUpdated` did.

orignal: It must be with interval as specified in this config parameter. Otherwise you might end up with empty netdb quickly

hakunamtu: Sorry for the long delay. As far as I can see, doing it less frequently allows netdb to bloat uncontrollably. One apparent issue that needs to be fixed is that recently updated (`IsUpdated() == true`) but now unreachable records are deleted from netdb, which deviates from existing behaviour. Otherwise netdb service does the same thing, except for less frequent disk writes. If there is something else I fail to see, please, tell me.

hakunamtu: Also persisting netDb on exit does not work with the default systemd unit's stop signal - SIGQUIT.

AVAtarMod:
Why this PR not closed yet? As I know, it was merged after 2.23?

orignal:
Потому что не фонтан

PR #1295, opened 2019-02-11
Support websocket connections over HTTP proxy

l-n-s:


l-n-s:
Original ticket: https://trac.i2p2.de/ticket/2422

libi2pd_client/HTTPProxy.cpp
```diff
@@ -219,7 +219,9 @@ namespace proxy {
 		/* replace headers */
 		req.UpdateHeader("User-Agent", "MYOB/6.66 (AN/ON)");
 		/* add headers */
-		req.UpdateHeader("Connection", "close"); /* keep-alive conns not supported yet */
+		auto h = req.GetHeader ("Connection");
+		if (h.find("upgrade") == std::string::npos && h.find("Upgrade") == std::string::npos)
```
orignal: Better to try to find "pgrade" and then compare lowercase of [-1] with 'u'

l-n-s: Something like that?

`if (std::tolower(h[x - 1], std::locale("C")) == 117) do_stuff();`

l-n-s: check now

Issue #1346, opened 2019-05-07
Don't use IPv4 from special address space as external IP


r4sas:
https://en.wikipedia.org/wiki/IPv4#Special-use_addresses

todo: check if the found external IP address is in these subnets, and treat them as invalid.
Generally:
* 10.0.0.0/8
* 100.64.0.0/10
* 127.0.0.0/8
* 172.16.0.0/12
* 192.168.0.0/16

add: https://github.com/i2p/i2p.i2p/blob/master/installer/resources/blocklist.txt

l-n-s:
It also needs an option to allow private networks for the testnet

Issue #1311, opened 2019-03-20
After I set ntcp=false and restart i2pd, it is still present in the router.info


l-n-s:

createS0ftware:
How are you able to read router.info ? Whilst debugging in code ? I can't get plain text out of it

orignal:
routerinfo from i2pd-tool

createS0ftware:
Thanks @orignal

r4sas:
NTCP was dropped, some changes in current code must fix issue. Open new issue if problem still exists.

Issue #1275, opened 2018-12-05
boost can't bind udp socket on (maybe) tun interface with yggdrasil on it


r4sas:
>Hi, I am having a problem that i2pd 2.21.1-1 stops working after a few reboots (or if day changes during poweroff?). Would this happen to be fixed on 2.22.0 which isn't in Debian repositories yet?

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=915501

_Originally posted by @Mikaela in https://github.com/PurpleI2P/i2pd/commit/779228857e5b212fbeff161dced60a61e1bc0c3e#commitcomment-31557786_

r4sas:
I'll check as soon as possible in sid.

l-n-s:
Not related, but when installing Ubuntu 18.04 ppa package there is a weird warning displayed:

```
Setting up libboost-program-options1.65.1:amd64 (1.65.1+dfsg-0ubuntu5) ...
Setting up libboost-filesystem1.65.1:amd64 (1.65.1+dfsg-0ubuntu5) ...
Setting up i2pd (2.22.0-1bionic1) ...
adduser: Warning: The home directory `/var/lib/i2pd' does not belong to the user you are currently creating.
Processing triggers for libc-bin (2.27-3ubuntu1) ...
Processing triggers for systemd (237-3ubuntu10.9) ...
```

LLE8:
Please show 
`grep i2pd /etc/passwd`
`ls -ld /var/lib/i2pd`

orignal:
LLE8 там еще openssl свинью подкладывает, пытаясь читать свой кофиг

LLE8:
Да, у меня в Stretch такое же предупреждение. 
Воспроизвести просто
1 `apt-get purge i2pd`
2 `userdel -r i2pd`
3 `dpkg -i <path_to_you_package>/i2pd_2.22.0-1_amd64.deb`, если есть собственный пакет или  `apt-get install i2pd`, если из репозитория
Однако как мне кажется, там нет ничего страшного, т.к. потом выполняется usermod с нужными изменениями. можно выполнить `apt-get remove  i2pd` и `dpkg -i <path_to_you_package>/i2pd_2.22.0-1_amd64.deb` и увидеть `usermod: no changes`.

r4sas:
Это нормально.
Но просьба не забивать иссуй посторонними сообщениями.

r4sas:
[...]

Issue #1251, opened 2018-10-07
RFC4941 IPV6 addresses shouldn't be published in netdb


madman2003:
i2pd picks IPV6 addresses with very short lifetime, so one or two temporary addresses when the hard expire occurs all IPV6 tunnels collapse. Later new IP is picked up again and published in netdb, and a little while later the problem same happens.

orignal:
It takes ipv6 address to publish from a successive NTCP connection.
Please provide an example.

madman2003:
Basically with IPV6 you have 128 bits address. The lower 64 bits vary within the same network. One address is link local, starting with FE80 (upper 16 bits), not very interesting for global routing. One address will be long term stable and globally routable. Optionally/often there is also at least one temporary address that is valid for X minutes (all addresses share the same upper 64 bits, but usually completely different lower 64 bits). OS prefers the temporary address for newly established outbound communication for privacy reasons, because stable address embeds MAC address in lower 64 bits (using the oldest spec for IPV6 address generation) allowing global tracking of that MAC address in any network (upper 64 bits). Old temporary privacy addresses stay around for a while as deprecated, but eventually disappear.

For IPV6 it is very common to have multiple IP addresses at any given time, and you can communicate with all of them. Checking  a random outbound connection for IP is not reliable for server purposes, because the address can expire within 5-10 minutes. Java I2P has some protections to select long term stable address, although it may be OS dependent.

l-n-s:
@orignal but what if we don't have SSU enabled?

Looks like OP is proposing to add some address  sanity checking, so that we don't add "temporary privacy addresses" to RouterInfo.

orignal:
if no SSU, somebody has to specify it explicitly

madman2003:
@l-n-s My suggestion was to do some sanity checking, but I'm here to raise awareness about a problem, not dictate solutions. I was just surprised to see i2pd not working well for IPV6, which I isolated to the router info containing an invalid ip, which previously had been a temporary address.

r4sas:
>20:15 <~R4SAS> https://tools.ietf.org/html/rfc4941
20:16 <~R4SAS> #section-3.3
20:18 <~R4SAS> https://stackoverflow.com/questions/35442119/find-temporary-ipv6-address-using-getadaptersaddresses
20:24 <~R4SAS> https://stackoverflow.com/questions/17833765/detect-temporary-ipv6-address-crossplatform
[...]

PR #1349, opened 2019-05-15
Transports, RI, socket binding code rewrite

r4sas:
Target is to cover issues specified to transports and RI.

- [x] #510 - done. Also, here can be more things to do, please report if any
- [ ] #837
- [ ] #1219
- [ ] #1248
- [x] #1251 - will work only on linux
- [ ] #1275 - RI and socket binding review
- [x] #1311 - checked, no more happens
- [x] #1346 - check executed on RI parsing
- [x] #1363 - lack of exceptions handlling...

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -6,7 +6,7 @@ project ( "i2pd" )
 # for debugging
 #set(CMAKE_VERBOSE_MAKEFILE on)
 
-# configurale options
+# configurable options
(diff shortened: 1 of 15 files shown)


r4sas:
seems like that kind of preference done a trick. need to test it on real machine with IP from security extensions.

r4sas:
Closing because big amount of changes done since this PR was opened.

Issue #1357, opened 2019-05-25
BOB fails to create inbound tunnel


rszibele:
The connection gets closed when you try to create an inbound tunnel and there is an error in the stdout/stderr of i2pd.

```
$ nc localhost 2827
BOB 00.00.10
OK
setnick test-source
OK Nickname set to test-source
newkeys
OK HnN2n3wK78pVRfNInYJsRMYwrOu-G7i1tFWIHIsnSpNzLltRdgDjbH3vMk0KkkaR0s0DX3r~P-U~NR6H5ekDx5SP2~qGPKc4KykWQttCmJkaY5gRc5mMtCviHL8Vx6ilgh0ttQ0PMhAsIDjPac2O8jK4ccqWF2ufS8IGOCEUKIyCNA8zq85-VeJH-6f0n~5jjihaMofaam7J31IOoZgREBoN9K70fTTjURrogNFoJXyZKC3qAnvhwjQmx12NdnwmwKZxRlKq7l421e6dE4Y7z~kshMiQizsYBEQimAQRzPw0gVYGwFe59fP0kZYa21xDjsFn82XxiKg9tkGr78LpcREs15uncN0ZmFE7NNBlrDzyT~SzTf4mfSBq7fcII30ePVHWiNNfo0R09oknyZF~-okQzqK1AV2qjxNxi7~JaDnNAsOAuu4g1TdIGjLho4uREbR35jcyeC0CYhfEzvHMmTTTCB-H3LTfCMdeDATFIbUxAN0HjnZNV-x5HCJ5jdtxAAAA
outhost localhost 
OK outhost set
outport 33555
OK outbound port set
start
```
i2pd output:
```09:20:16@223/error - BOB: runtime exception: Invalid argument```

I also created an outbound tunnel and tried to connect to it but it fails with destination not found.

rszibele:
caused by:
https://github.com/PurpleI2P/i2pd/blob/07405e57b97865e2b68f6411b27ffb67e7079149/libi2pd_client/BOB.cpp#L206

from_string docs:
> Create an address from an IPv4 address string in dotted decimal form, or from an IPv6 address in hexadecimal notation.

A resolver would need to be used to support this:
https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__basic_resolver/resolve.html

gdb backtrace:
```gdb
[...]

Issue #1355, opened 2019-05-24
BOB segmentation fault when tunnel settings are incomplete


rszibele:
Opening a tunnel which allows incoming connections causes a segmentation fault:

Start i2pd with bob support:
```
$ /opt/i2pd/bin/i2pd --bob.enabled=1
```

Connect to bob and enter the commands:
```
$ nc localhost 2827
BOB 00.00.10
OK
setnick test-source
OK Nickname set to test-source
outhost localhost
OK outhost set
outport 33556
OK outbound port set
quiet True
OK Quiet set
start
```

GDB backtrace:
```gdb
Thread 15 "i2pd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe3fff700 (LWP 27571)]
__memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:70
70      ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S: No such file or directory.
(gdb) bt
#0  __memcmp_avx2_movbe () at ../sysdeps/x86_64/multiarch/memcmp-avx2-movbe.S:70
#1  0x0000555555c60f28 in i2p::data::Tag<32ul>::operator< (this=0x5555561ecb50, other=...)
    at /home/richard/Documents/i2pd/build/../libi2pd/Tag.h:32
#2  0x0000555555c677c1 in std::less<i2p::data::Tag<32ul> >::operator() (this=0x5555561b8628 <i2p::client::context+40>, __x=..., 
    __y=...) at /usr/include/c++/8/bits/stl_function.h:386
#3  0x0000555555ef8bcc in std::_Rb_tree<i2p::data::Tag<32ul>, std::pair<i2p::data::Tag<32ul> const, std::shared_ptr<i2p::client::ClientDestination> >, std::_Select1st<std::pair<i2p::data::Tag<32ul> const, std::shared_ptr<i2p::client::ClientDestination> > >, std::less<i2p::data::Tag<32ul> >, std::allocator<std::pair<i2p::data::Tag<32ul> const, std::shared_ptr<i2p::client::ClientDestination> > > >::_M_lower_bound (this=0x5555561b8628 <i2p::client::context+40>, __x=0x5555561ecb30, __y=0x5555561b8630 <i2p::client::context+48>, __k=...)
    at /usr/include/c++/8/bits/stl_tree.h:1888
[...]

Issue #1352, opened 2019-05-22
BOB list command has wrong output format


rszibele:
With the `list` command i2pd only outputs the nickname and without a `:` after the key so it's not compatible with the Java I2P reference implementation.

```
$ nc localhost 2827
BOB 00.00.10
OK
list
NICKNAME test
OK Listing done
```

It should be:
```
list
DATA NICKNAME: test STARTING: false RUNNING: true STOPPING: false KEYS: true QUIET: false INPORT: 33555 INHOST: localhost
OK Listing done
```

orignal:
will fix

rszibele:
Will the other keys and values `STARTING`, `RUNNING`, `STOPPING`, `QUIET`, `INPORT` and `INHOST` also be added?

orignal:
It reequires more work.  Maybe later. Use 'status' for now

Issue #1351, opened 2019-05-21
i2pd BOB implementation doesn't accept some commands from telnet


rszibele:
If you connect to BOB via. telnet like described here https://geti2p.net/en/docs/api/bob then it doesn't work as expected. Some commands work, but others do not:

![image](https://user-images.githubusercontent.com/1387224/58061144-327db500-7b75-11e9-89f2-0b4a4d508480.png)

![image](https://user-images.githubusercontent.com/1387224/58061531-d2880e00-7b76-11e9-8f39-80852f60ca47.png)
![image](https://user-images.githubusercontent.com/1387224/58061634-4f1aec80-7b77-11e9-9f9d-3e368534f429.png)

I'm working on a BOB C++ (QTcpSocket) client library and I use `\n` terminators and all commands work. Here's a Wireshark grab:
![image](https://user-images.githubusercontent.com/1387224/58061365-27775480-7b76-11e9-816f-908169dd9515.png)
![image](https://user-images.githubusercontent.com/1387224/58061413-4ece2180-7b76-11e9-8beb-1cd27d04293a.png)

The only difference between telnet and my working library are the line terminators `\r\n` vs `\n`. I assume it's due to `help`, `start` and `newkeys` requiring no arguments while `setnick`, `inhost` and `inport` require an argument, so it's a different code path where `\r\n` isn't handled.

r4sas:
https://github.com/PurpleI2P/i2pd/blob/60ec03237e61d0cec24feacb68abacdf872cbb27/libi2pd_client/BOB.cpp#L256-L260

orignal:
You must use nc instead

LLE8:
May be helpful
```
man telnet
....
                crlf          If this is TRUE, then carriage returns will be
                              sent as <CR><LF>.  If this is FALSE, then car‐
                              riage returns will be send as <CR><NUL>.  The
                              initial value for this toggle is FALSE.
```

rszibele:
Is this by design or should i2pd BOB be compatible with Java I2P BOB reference implementation?

orignal:
I guess Java just replaces "\r\n" to "\n" when read next string

majestrate:
On Tue, May 21, 2019 at 06:13:44AM -0700, orignal wrote:
> I guess Java just replaces "\r\n" to "\n" when read next string
is BOB properly buffered or does it require the line sent in the same tcp segment?

rszibele:
[...]

PR #1359, opened 2019-05-29
BOB modifications.

rszibele:
Hi, I've fixed/added/changed the following:

1. BOBCommandSession: added telnet support (https://github.com/PurpleI2P/i2pd/issues/1351) and removed C string pointer arithmetic.
2. Cleaned up the status command.
3. Added the other outputs to the `list` command (https://github.com/PurpleI2P/i2pd/issues/1352#issuecomment-494902499) and it now uses the same string building function like the status command.
4. Added extra validation to the `start` command. It prevents a segfault (https://github.com/PurpleI2P/i2pd/issues/1355) and also prevents some inconsistent states.
5. Added the `help` command.
6. Separated the address into outhost and inhost (will be required for resolving).

**Future TODO:**
If this patch is accepted and I have some time I might implement host resolving if you want to add it (https://github.com/PurpleI2P/i2pd/issues/1357). Java BOB supports arbitrary hostnames, e.g. you can create an outbound tunnel to google.com.

Is there an IRC channel or mailing list for i2pd?

### DIFF
--- a/libi2pd_client/BOB.cpp
+++ b/libi2pd_client/BOB.cpp
@@ -50,7 +50,7 @@ namespace client
 	void BOBI2PInboundTunnel::ReceiveAddress (std::shared_ptr<AddressReceiver> receiver)
 	{
 		receiver->socket->async_read_some (boost::asio::buffer(
-		        receiver->buffer + receiver->bufferOffset,
+				receiver->buffer + receiver->bufferOffset,
 				BOB_COMMAND_BUFFER_SIZE - receiver->bufferOffset),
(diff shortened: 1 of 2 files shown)


K1773R:
@rszibele there is #i2pd-dev on irc2p.

orignal:
People complain, one of these changes has broken transmission-i2p. Must be investigated.

rszibele:
I will look into it. Is there a specific error message?

orignal:
I asked for the details. But I guess it's easier to just run transmission and try..

PR #1374, opened 2019-06-25
BOB: fix status command.

rszibele:
This PR fixes the status command, in which the response now correctly starts with "OK DATA" and prefers to use the destination tunnel data if it is created.

I tried it with transmission-i2p and it now correctly identifies that the tunnel is created and running. I however could not test it beyond that, because transmission-i2p always has a tracker error (with old i2pd version too). It would be nice if someone who uses transmission-i2p regularly can test this as well.

This fixes the transmission-i2p problem mentioned in: https://github.com/PurpleI2P/i2pd/pull/1359 .

### DIFF
--- a/libi2pd_client/BOB.cpp
+++ b/libi2pd_client/BOB.cpp
@@ -341,10 +341,10 @@ namespace client
 		SendReplyOK();
 	}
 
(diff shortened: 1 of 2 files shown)


orignal:
Thanks. I will let them know and ask for testing

PR #1395, opened 2019-07-24
Improve i2pd.conf example

zetok:
- In case of default port setting, change it to `0` so that a random one
  would be assigned for i2pd in case where user uncomments the setting
  without changing the port
- In case of ipv6 setting change it to `true` since the default
  behaviour is to not use IPv6, and the only reason to uncomment it
  would be to set it to `true`.
- try to specify default values and/or default behaviour
- make things more consistent
- remove trailing whitespaces
- formatting

### DIFF
--- a/contrib/i2pd.conf
+++ b/contrib/i2pd.conf
@@ -1,231 +1,337 @@
-## Configuration file for a typical i2pd user
-## See https://i2pd.readthedocs.io/en/latest/user-guide/configuration/
-## for more options you can use in this file.
+#   Configuration file for a typical i2pd user
+#   See https://i2pd.readthedocs.io/en/latest/user-guide/configuration/
+#   for more options you can use in this file.
 
-## Lines that begin with "## " try to explain what's going on. Lines
-## that begin with just "#" are disabled commands: you can enable them
-## by removing the "#" symbol.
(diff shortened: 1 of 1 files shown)


l-n-s:
Why did you remove double # before verbose description comments? It was made like that for a reason.

zetok:
To make it more readable and consistent with most config files. I guess I should add that to the commit message?

```
$ find /etc -name '*.conf' -exec grep -l '^## ' {} + 2>/dev/null | wc -l
9
$ find /etc -name '*.conf' -exec grep -l '^# ' {} + 2>/dev/null | wc -l
110
```

l-n-s:
I personally don't see how it improves readability and consistency. Nobody ever complained.

zetok:
>Nobody ever complained.

If this isn't clear enough, the very change in the PR is a complaint and a fix of the complaint, all in one.

>I personally don't see how it improves readability and consistency.

I'd suggest you to look at my previous comment again then, since I've provided sample data and method of obtaining the data supporting the claim of improved consistency.

Regarding readability, you don't improve it by adding/having things that are unnecessary, but by removing unnecessary things and adding whitespaces (not the character, just visually empty space between things) wherever needed.

PR #1409, opened 2019-08-24
android - show battery optimiz. os dialog menu item added; translated all battery stuff into Russian

nonlinear-chaos-order-etc:


nonlinear-chaos-order-etc:
i currently do some more changes, so please wait with merging

PR #1453, opened 2020-01-12
Update Dockerfile

donarrock:
Fixes dependencies, re-adds `boost-python` as `boost-python2` which was removed in #1408 .


r4sas:
@donarrock can you recheck if something changed since opening PR?
Due to @l-n-s inactivity I'll merge it after this.

donarrock:
I can confirm it works for me with 2.30.0 and 2.31.0, so I think it should be fine.

r4sas:
Ok, thanks.

PR #1459, opened 2020-01-20
libi22pd: convert several loops to range based ones

neheb:
Found with clang-tidy's modernize-loop-convert.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


orignal:
I doubt it's faster than C-style two-indices array, e.g., say table[i][j] is just one assembly instruction.
That's C++ rather than Pythion.

PR #1460, opened 2020-01-20
libi2pd: use std::move and pass by value

neheb:
Found with clang-tidy's modernize-pass-by-value

Signed-off-by: Rosen Penev <rosenp@gmail.com>


orignal:
RouterInfo (std::string fullPath);

You can't declare copy constructor this way

orignal:
No worth for it

PR #1477, opened 2020-02-27
drop websockets support

r4sas:
As discussed before on Ilita IRC, we think about dropping websockets support due to lack of usage (we've never seen any kind of its usage).

Opening PR for discussion about that idea.


alopatindev:
I wish there was WebSocket support in i2pd. This would make [Nostr](https://en.wikipedia.org/wiki/Nostr) (and things like [ZapThreads](https://github.com/codonaft/zapthreads-codonaft#readme)) work.

Nostr is currently available on [tor onion](https://github.com/0xtrr/onion-service-nostr-relays#readme) services besides [clearnet](https://github.com/aljazceru/awesome-nostr#relay-lists), but not on i2p.

PR #1478, opened 2020-03-01
add ntcp2proxy support

unlnown542a:
adding ntcp2proxy support

### DIFF
--- a/libi2pd/NTCP2.cpp
+++ b/libi2pd/NTCP2.cpp
@@ -20,19 +20,21 @@
 #include "Transports.h"
 #include "NetDb.hpp"
 #include "NTCP2.h"
+#include "HTTP.h"
+#include "util.h"
 
 namespace i2p
 {
 namespace transport
 {
 	NTCP2Establisher::NTCP2Establisher ():
-		m_SessionRequestBuffer (nullptr), m_SessionCreatedBuffer (nullptr), m_SessionConfirmedBuffer (nullptr) 
-	{ 
+		m_SessionRequestBuffer (nullptr), m_SessionCreatedBuffer (nullptr), m_SessionConfirmedBuffer (nullptr)
+	{
 	}
 
-	NTCP2Establisher::~NTCP2Establisher () 
-	{ 
-		delete[] m_SessionRequestBuffer; 
+	NTCP2Establisher::~NTCP2Establisher ()
+	{
+		delete[] m_SessionRequestBuffer;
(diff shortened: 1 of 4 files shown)


libi2pd/NTCP2.cpp
```diff
@@ -371,10 +373,82 @@ namespace transport
...
+			memcpy(buff+4, addrbytes.data(), addrsize);
+		}
+		else if (addrtype == eHostname)
+		{
+			buff[3] = 0x03;
+			size_t addrsize = host.size();
+			sz = addrsize + 1 + 4;
+			if (2 + sz > sizeof(buff))
+			{
+				// too big
```
majestrate: maybe you should put a log message here about the host name being too big?

libi2pd/NTCP2.cpp
```diff
@@ -1094,14 +1168,158 @@ namespace transport
...
+		}
+		else
+			LogPrint(eLogError, "NTCP2: unknown proxy type, invalid state");
+	}
+
+	void NTCP2Server::ConnectWithProxy (const std::string& host, uint16_t port, RemoteAddressType addrtype, std::shared_ptr<NTCP2Session> conn)
+	{
+		if(m_ProxyEndpoint == nullptr)
+		{
+			return;
```
majestrate: perhaps make this an assert?

libi2pd/NTCP2.h
```diff
@@ -248,9 +271,15 @@ namespace transport
...
 			std::unique_ptr<boost::asio::ip::tcp::acceptor> m_NTCP2Acceptor, m_NTCP2V6Acceptor;
-			std::map<i2p::data::IdentHash, std::shared_ptr<NTCP2Session> > m_NTCP2Sessions; 
+			std::map<i2p::data::IdentHash, std::shared_ptr<NTCP2Session> > m_NTCP2Sessions;
 			std::list<std::shared_ptr<NTCP2Session> > m_PendingIncomingSessions;
 
+			ProxyType m_ProxyType;
+			std::string m_ProxyAddress;
+			uint16_t m_ProxyPort;
+			boost::asio::ip::tcp::resolver m_Resolver;
+			boost::asio::ip::tcp::endpoint * m_ProxyEndpoint;
```
majestrate: if it's not owned by anyone else using `std::unique_ptr<boost::asio::ip::tcp::endpoint>` would be easier here. (maybe).

orignal: I will take care. At least his code works

PR #1535, opened 2020-06-13
Improve AppArmor profile

komachi:
- give it a name
- import needed abstractions (on debian stable I also get denials for /etc/ssl/openssl.cnf, so openssl abstraction is imported, most of profile contained duplicated rules from nameservice abstraction, importing named abstraction also fixes all problems like #1192 which are likely to appear on some setups)
- allow local additions
- cleanup

### DIFF
--- a/contrib/apparmor/usr.sbin.i2pd
+++ b/contrib/apparmor/usr.sbin.i2pd
@@ -4,34 +4,22 @@
 #
 #include <tunables/global>
 
-/usr/sbin/i2pd {
+profile i2pd /{usr/,}sbin/i2pd {
   #include <abstractions/base>
-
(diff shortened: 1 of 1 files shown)


r4sas:
Hello. Thank you for imrovements!
I have question about back-compatability: is `openssl` and `nameservice` abstractions available on earliear versions of apparmor, as example on Jessie, Stretch and corresponding versions on Ubuntu?

komachi:
They are part of core apparmor package for years, so yes, they should be available everywhere, [even on jessie](https://packages.debian.org/jessie/amd64/apparmor/filelist).

r4sas:
In that case let's merge it ;) Thank you!

PR #1553, opened 2020-09-18
Hardening and more for the systemd service

rany2:
Performance improvements, auto-restart on failure to prevent unattended routers from shutting down by accident, and leveraging systemd's security features.


contrib/i2pd.service
```diff
@@ -2,31 +2,60 @@
...
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectSystem=strict
+RestrictAddressFamilies=AF_INET AF_INET6
+RestrictNamespaces=cgroup ipc mnt pid user uts
+RestrictRealtime=true
+RestrictSUIDSGID=true
+RemoveIPC=true
+ReadOnlyDirectories=/
+ReadWriteDirectories=-/proc
```
majestrate, line 54: why read and write to `/proc`? shouldn't this be at most read only?

analotia: It **_seems_** like it works fine with `/proc` as read only. However I don't see why not allow it full access to `/proc`? 

After all it's going to be able to write in `/proc`  to it's own i2p user only, so read-only and read-write wouldn't make any meaningful difference security-wise.  The plus is that we minimize chances of something going wrong.

majestrate: full access to `/proc` is similar in nature to full access to `/sys` both can be abused to possibly exploit a theoretical privesc bug. also general practice is to allow only what is needed and nothing more.

analotia: `/proc` line got removed.

analotia: Is there an option to disable daemon? Something like `--no-daemon`?

Older default debian configurations had `daemon=true` but thankfully this is no longer default.

analotia: This doesn't seem true. My bad

analotia:
We might want to change loglevel to error in the systemd service.

majestrate:
> We might want to change loglevel to error in the systemd service.

i agree.

analotia:
done, I'm not sure if anyone anyone prefers a better way to do this

contrib/i2pd.service
```diff
@@ -2,31 +2,59 @@
...
 Group=i2pd
+PermissionsStartOnly=yes
+
 RuntimeDirectory=i2pd
 RuntimeDirectoryMode=0700
 LogsDirectory=i2pd
 LogsDirectoryMode=0700
-Type=forking
-ExecStart=/usr/sbin/i2pd --conf=/etc/i2pd/i2pd.conf --tunconf=/etc/i2pd/tunnels.conf --tunnelsdir=/etc/i2pd/tunnels.conf.d --pidfile=/run/i2pd/i2pd.pid --logfile=/var/log/i2pd/i2pd.log --daemon --service
+Type=exec
```
majestrate, line 16: isn't the type called `simple` not `exec` ? or is there a difference here?

analotia: from `man 5 systemd.service`: 
>simple proceeds with further jobs right after fork() returns, while exec will not proceed before both fork() and execve() in the service process succeeded. Note that this means systemctl start command lines for exec services will report failure when the service's binary cannot be invoked successfully (for example because the selected User= doesn't exist, or the service binary is missing).

analotia: to me that made it seem like exec was the better choice but it probably doesn't matter too much  

i don't mind making it simple

majestrate:
this looks good so far to me.

majestrate (approved):

analotia:
some more hardening after using `systemd-analyze security i2pd`.

These changes shouldn't cause any issues

analotia:
@majestrate check the latest version of the service file. Please  verify that it still should be approved

analotia:
it's working fine so far

r4sas:
I won't apply changes without back-compatability patches for systems which uses systemd (most of them). Yes, I know that this a bit strange, but such hardening and using all available options always broke support with systems which doesn't have them.

contrib/i2pd.service
```diff
@@ -2,31 +2,64 @@
...
 
 KillSignal=SIGQUIT
 # If you have the patience waiting 10 min on restarting/stopping it, uncomment this.
 # i2pd stops accepting new tunnels and waits ~10 min while old ones do not die.
 #KillSignal=SIGINT
 #TimeoutStopSec=10m
 
-# If you have problems with hanging i2pd, you can try increase this
-LimitNOFILE=4096
+LimitNOFILE=65536
```
r4sas, line 30: Too much, don't think that you can overrun 8K.

PR #1575, opened 2020-11-02
Change the enableuniquelocal to false by default to improve anonimity

### LINKED ISSUE #1573: Users are identifiable if server is running i2pd
If you, somehow, print the REMOTE_ADDR environment variable, it will print a bogon ip address (127.x.x.x)
however, this bogon IP is different for every client (meaning a client has 127.x.y.z and another client has 127.y.x.z), java i2p set the remote IP address as 127.0.0.1 for everyone

qorg11:
Fixes #1573


majestrate:
i dont understand how this improves anonymity by default?

orignal:
Current code works the same way as Java-I2P.

qorg11:
>i dont understand how this improves anonymity by default?

It is more anonymous if everyone has the same IP address

>Current code works the same way as Java-I2P.

Really? Java-I2P set the IP address as 127.0.0.1 for what i've tested. (In latest version at least)

orignal:
>It is more anonymous if everyone has the same IP address
you can always set this parameter in YOUR config, but for whatever reason want to change DEFAULT value and break existing server functionality, because YOU think it's more anonymous.  

> Java-I2P set the IP address as 127.0.0.1 for what i've tested

I will check with zzz. Also it depends on OS.

PR #1623, opened 2021-01-21
Fix clang warning

rex4539:
Fixed the following warning:
```
libi2pd/NTCP2.cpp:1522:7: warning: lambda capture 'this' is not used [-Wunused-lambda-capture]
                                        [this, readbuff, timer, conn] (const boost::system::error_code & ec, std::size_t transferred)
                                         ^~~~~
1 warning generated.
```


rex4539:
@orignal

orignal:
Please mind gcc 4.8, your change might not be buildable there

PR #1626, opened 2021-01-27
make floodfill enabled by default

johnrosen1:
since it is year of 2021,most of devices run i2pd should have more than enough compute power to enable this feature,maybe we should consider enable it by default?


orignal:
You can't be a floodfill if you are behind the firewall. You can't be a floodwill if you don't have anough bandwidth.
You might not want to give too much CPU to i2pd.

PR #1640, opened 2021-03-07
logging opimization

brain5lug:
This change will make logging faster.
This will eliminate one string coping on each logging code line for code prior c++20 and two sring copies after c++20

The reason is the following:
For https://en.cppreference.com/w/cpp/string/basic_string/basic_string
you get a copy with small differences for pre C++20 and C++20.

The move around std::stringsteam is used to use proper `str() &&` function call.
For under C++ 20 version the same `str()` is called even with move.


r4sas:
Checking at Fedora Copr: https://copr.fedorainfracloud.org/coprs/supervillain/i2pd/build/2061380/

PR #1669, opened 2021-06-23
C wrapper for part of the libi2pd api

eyedeekay:
This creates a C wrapper for part of the libi2pd API, creates a [SWIG](http://www.swig.org) wrapper for that C API. and makes it usable from the standard Go toolchain(i.e. `go get`).  This should open up the possibility of embedding i2pd into any language which can be supported by SWIG, Aside from Go, where I'll be using it, there are Python, Ruby, and node.js applications which could benefit from it as well.

The parts of it I'm checking in are just the simple parts right now. I could not figure out how to wrap the Destination and Streaming classes yet, so I left those off. If I'm understanding it correctly, I would need to wrap each of these in a struct, then use capi.cpp to use the bare pointer to the struct to instantiate the shared_ptr to the class to actually use those functions? I'll keep working on it and send a PR for those at a later date.

While I was working on it I needed to statically compile libi2pd and I saw that the Makefile wasn't letting me do `CXXFLAGS=-static LDFLAGS=-static USE_STATIC=yes make` to create a static executable, so I changed the Makefile so that lets one pass `USE_*` options as environment variables. I also thought I would need the languages statically compiled into their own archive and added that capability to the Makefile, but this turned out to not be the case at this time. I left the target to compile the languages by themselves in the Makefile in case it's needed in the future.


r4sas:
Maybe better move wrapper sources to other directory, like `libi2pd_wrapper` and then link it with `libi2pd.a`?

eyedeekay:
> Maybe better move wrapper sources to other directory, like `libi2pd_wrapper` and then link it with `libi2pd.a`?

Sure I can do that

PR #1679, opened 2021-07-30
libi2pd: mark additional ipv6 addresses/nets as reserved

simonvetter:
This adds :: (undefined address), ::1 (loopback address) as well as
ff00::/8 (multicast prefix) to reservedIPv6Ranges.

A bunch of nodes seem to be publishing bogus addresses (mostly ::1)
in the netDB, resulting in unnecessary tunnel build failures.


r4sas:
Adding `::1` can broke following work with next detection of real IPv6, so I don't think that's good idea add it to that list...

simonvetter:
On ipv6-only nodes (no ipv4 address, ipv4 connectivity through a nat64 gateway which i2pd doesn't use), the tunnel creation success rate went from ~30% to around 40-50% and the number of active transit tunnels is about twice as high as before.

I went through every call to IsInReservedRange() and failed to see where it could be a problem, but you know this codebase better than I do.
In any case, my nodes don't seem to have issue detecting v6 connectivity.

r4sas:
Re-Merged as b3e7b1b5ac79e2283835a07e15efab3f9083756e

PR #1687, opened 2021-08-23
[tests] fix compilation of test-blinding

dbermond:
`test-blinding` currently fails to build with the following error:

```
In file included from ../libi2pd/Timestamp.cpp:19:
../libi2pd/RouterContext.h:21:10: fatal error: I18N_langs.h: No such file or directory
   21 | #include "I18N_langs.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

Adding headers search path for `i18n` directory fixes this.

How to reproduce:

```
$ git clone https://github.com/PurpleI2P/i2pd.git
$ make -C i2pd/tests
```

System: Arch Linux x86_64 with gcc 11.1.0 and boost 1.76.0.


r4sas:
Thank you for finding problem.
Your patch will make able to compile that test, but won't make it run on some systems due to lack of symbols.

r4sas:
I'll pull proposed changes with complete fixing of problem and commit them.

dbermond:
@r4sas Thank you for working on this.

PR #1698, opened 2021-10-18
Makefile.osx: detect openssl@1.1 from Homebrew

xanoni:
The current Makefile does not work on Darwin if openssl was installed
via Homebrew ... this change ensures the openssl headers and libs are
included if they were installed via Homebrew, even when i2pd is compiled
without Homebrew (e.g., by just cloning the repo and running `make`).

### DIFF
--- a/Makefile.osx
+++ b/Makefile.osx
@@ -1,10 +1,16 @@
 CXX = clang++
 CXXFLAGS := ${CXX_DEBUG} -Wall -std=c++11 -DMAC_OSX
-INCFLAGS = -I/usr/local/include
+INCFLAGS := -I/usr/local/include
 LDFLAGS := -Wl,-rpath,/usr/local/lib -L/usr/local/lib
 LDFLAGS += -Wl,-dead_strip
 LDFLAGS += -Wl,-dead_strip_dylibs
 LDFLAGS += -Wl,-bind_at_load
+
+$(shell brew ls openssl@1.1 > /dev/null 2>&1)
+ifeq ($(.SHELLSTATUS),0)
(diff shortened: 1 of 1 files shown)


r4sas:
RTFM: https://i2pd.readthedocs.io/en/latest/devs/building/unix/#mac-os-x
`make HOMEBREW=1`

xanoni:
Just reproduced the error, this is what I saw yesterday and why I sent this PR. It was trying to install something into my root folder but it obviously doesn't have permission to do that. That's why I compiled without `HOMEBREW=1` and made the change in this PR instead.

@r4sas how to avoid the below? set a custom prefix of `/usr/local`?

```bash
clang++ -o i2pd  -L/usr/local/opt/openssl@1.1/lib -L/usr/local/opt/boost/lib obj/daemon/Daemon.o obj/daemon/HTTPServer.o obj/daemon/I2PControl.o obj/daemon/UPnP.o obj/daemon/UnixDaemon.o obj/daemon/i2pd.o libi2pd.a libi2pdclient.a libi2pdlang.a -lz -lcrypto -lssl -lboost_system -lboost_date_time -lboost_filesystem -lboost_program_options -lpthread
install -d /bin /etc/i2pd /share/doc/i2pd /share/i2pd /share/man/man1 /var/lib/i2pd
install: chmod 755 /bin: Operation not permitted
install: mkdir /etc/i2pd: Permission denied
gmake: *** [Makefile.homebrew:42: install] Error 71
```

xanoni:
@r4sas please see https://github.com/PurpleI2P/i2pd/pull/1700

PR #1700, opened 2021-10-19
Makefile: set default target ("all")

xanoni:
Fixes compilation on Darwin, see:
https://github.com/PurpleI2P/i2pd/pull/1698#issuecomment-946304938

(make was incorrectly calling the "install" target instead of the "all" target)


xanoni:
I also updated the documentation: https://github.com/PurpleI2P/i2pd_docs_en/pull/57

PR #1703, opened 2021-10-31
memory leak and concurrency fixes

simonvetter:
This fixes a few memory leaks (destructors not being virtual) and concurrency issues (missing locks) in Tunnel, TunnelPool and TunnelConfig classes.

### DIFF
--- a/libi2pd/Tunnel.cpp
+++ b/libi2pd/Tunnel.cpp
@@ -366,6 +366,7 @@ namespace tunnel
 
 	std::shared_ptr<TunnelBase> Tunnels::GetTunnel (uint32_t tunnelID)
 	{
+		std::unique_lock<std::recursive_mutex> l(m_TunnelsMutex);
 		auto it = m_Tunnels.find(tunnelID);
 		if (it != m_Tunnels.end ())
 			return it->second;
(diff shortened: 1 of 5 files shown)


orignal:
Will slow down everything. Most of code is called from the single  tunnels thread.

simonvetter:
Reworked the code, added extra locks.

I haven't seen much of a slowdown in my tests but admittedly I've only been testing for a few hours.
At any rate, it fixes a lot of races reported by the thread sanitizer and a bunch of memory corruption issues. If the thread sanitizer is to be trusted, this code is definitely called from multiple concurrent threads.

r4sas:
That PR brokes usage on Windows.

simonvetter:
Would you have any logfile, core dump or pointer so I can debug that windows deadlock? I'd like to fix it and re-submit the PR if possible.
Unfortunately I don't have any windows system around to test.

simonvetter:
By the way, if you'd rather fix races and memory issues yourself, I can post the threadsan/asan logs in here. Let me know.

r4sas:
> Would you have any logfile, core dump or pointer so I can debug that windows deadlock? I'd like to fix it and re-submit the PR if possible. Unfortunately I don't have any windows system around to test.

It's really hard to debug such things on windows because we are using MinGW g++ and here no such debuggers like in MSVS. On start no dumps created inside gdb and no log lines about errors in execution inside tunnels thread.

I'll try to add print to logs in places where you made changes, but it will be later (lack of time).

PR #1745, opened 2022-03-20
Setting up some response to file descriptor exhaustion

r4sas:
Proposed commit from user *det* on Irc2P.


orignal:
number of descriptor also depends on number of timer. That means SSU and Streaming

PR #1752, opened 2022-04-29
Added DESTDIR: Support for Staged Installs

CurtisLeeBolin:
reference: https://www.gnu.org/prep/standards/html_node/DESTDIR.html


r4sas:
How that variable differs from `PREFIX`?

PR #1769, opened 2022-06-25
Add SIGTSTP, SIGCONT support

AVAtarMod:
I added support SIGTSTP, SIGCONT signal for UNIX daemon. May be I created commits not right, because of I want do things above:

1. First commit contain support "only signal catching" without config option
2. Second commit add signal catching as option

### DIFF
--- a/daemon/UnixDaemon.cpp
+++ b/daemon/UnixDaemon.cpp
@@ -60,7 +60,7 @@ void handle_signal(int sig)
             i2p::transport::transports.SetOnline(false);
             break;
         case SIGCONT:
-            LogPrint(eLogInfo, "Daemon: Got SIGCONT, restore connection to network...");
+            LogPrint(eLogInfo, "Daemon: Got SIGCONT, restoring connection to network...");
             i2p::transport::transports.SetOnline(true);
             break;
         }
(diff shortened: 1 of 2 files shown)


daemon/UnixDaemon.cpp
```diff
@@ -54,7 +55,15 @@ void handle_signal(int sig)
...
-	}
+        case SIGTSTP:
+            LogPrint(eLogInfo, "Daemon: Got SIGTSTP, disconnecting from network...");
+            i2p::transport::transports.SetOnline(false);
+            break;
+        case SIGCONT:
+            LogPrint(eLogInfo, "Daemon: Got SIGCONT, restoring connection to network...");
+            i2p::transport::transports.SetOnline(true);
+            break;
+        }
```
AVAtarMod, line 66: Log messages may be bad and unclean, because of my English is bad.

daemon/UnixDaemon.cpp
```diff
@@ -171,19 +180,27 @@ namespace i2p
...
+            sigaction(SIGUSR1, &sa, 0);
+            sigaction(SIGABRT, &sa, 0);
+            sigaction(SIGTERM, &sa, 0);
+            sigaction(SIGINT, &sa, 0);
+            sigaction(SIGPIPE, &sa, 0);
+            if (handleTSTP)
+            {
+                sigaction(SIGTSTP, &sa, 0);
+                sigaction(SIGCONT, &sa, 0);
+            }
```
AVAtarMod, line 201: I recommend replace 0 to nullptr, this is more readable:
```suggestion
            sigaction(SIGHUP, &sa, nullptr);
            sigaction(SIGUSR1, &sa, nullptr);
            sigaction(SIGABRT, &sa, nullptr);
            sigaction(SIGTERM, &sa, nullptr);
            sigaction(SIGINT, &sa, nullptr);
            sigaction(SIGPIPE, &sa, nullptr);
            if (handleTSTP)
            {
                sigaction(SIGTSTP, &sa, nullptr);
                sigaction(SIGCONT, &sa, nullptr);
            }
```

libi2pd/Config.cpp
```diff
@@ -311,6 +311,19 @@ namespace config {
...
+#ifdef __linux__
+        options_description unix_specific("UNIX-specific options");
+        unix_specific.add_options()
+            ("unix.handle_sigtstp", bool_switch()->default_value(false),
+                    "Switch to offline mode if received signal TSTP (SIGTSTP)"
+                    "(you   can send it by pressing CTRL+Z in terminal or with"
+                    " help commapnd kill and others, a.e pkill, if i2pd in "
+                    "daemon mode). If you need to switch to online mode, send"
+                    " signal CONT (SIGCONT)")
+        ;
```
AVAtarMod, line 323: As I said above, message need to rewrite to make it more clean and readable

libi2pd/Config.cpp
```diff
@@ -334,6 +347,9 @@ namespace config {
 			.add(persist)
 			.add(cpuext)
 			.add(meshnets)
+#ifdef __linux__
```
AVAtarMod, line 350: May be LINUX is more readable, we can define it in some main header as
```cpp
#ifdef __linux__
#define LINUX
#endif
```

AVAtarMod:
If you want, also I can sign commits with my GPG cert

PR #1786, opened 2022-08-29
Add support mac m1 and fix homebrew paths

d351d3r:
In new Macbooks, homebrew has the /opt/homebrew patch, not /usr/local/


r4sas:
Please do not mess-up fixed with whole project code reformat.

PR #1794, opened 2022-10-13
Update HTTPProxy.cpp

jiigen:
Remove DoNotTrack flag from HTTP Request header.

### DIFF
--- a/libi2pd_client/HTTPProxy.cpp
+++ b/libi2pd_client/HTTPProxy.cpp
@@ -249,6 +249,7 @@ namespace proxy {
 		req.RemoveHeader("Via");
 		req.RemoveHeader("From");
 		req.RemoveHeader("Forwarded");
+		req.RemoveHeader("DNT"); // Useless DoNotTrack flag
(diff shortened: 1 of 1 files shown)


r4sas:
Why remove this line? This flag makes no sense.

jiigen:
> Why remove this line? This flag makes no sense.

Yes...it's useless so why keep it?

r4sas:
Because in case of leaving it in requests it can be used for fingerprinting (two possible values with and w/o flag).

jiigen:
Exactly, maybe there's a misunderstanding. I've added that line not removed it. The flag DNT is actually not removed by i2pd http proxy.

r4sas:
Sorry, overworked at work :)

In this case, we must make sure to add the same in Java I2P so that clients do not differ for services.

jiigen:
> Sorry, overworked at work :)
> 
> In this case, we must make sure to add the same in Java I2P so that clients do not differ for services.

I agree but currently they already differ in the HTTP Headers anyway. Java I2P doesn't remove Accept* while i2pd removes them all and only keeps "Accept-Encoding" (which is better IMHO)

PR #1795, opened 2022-10-16
Add SAM UDP port param and fix for typo in config

polistern:
## SAM UDP port parameter

Selecting UDP port for listening based in TCP port (`port - 1`) is a pretty ambiguous.  
Added a separate backwards compatible option to specify UDP port in the config file.

```ini
[sam]
enabled = true
address = 127.0.0.1
port = 7656
portudp = 7655
```

## SSU option typo

Current code for this option is:

```c++
("ssu", bool_switch()->default_value(false), "Enable SSU transport (default: disabled)")
```

So typo is fixed from:

```ini
## Enable SSU transport (default = true)
ssu = false
```

To:

```ini
## Enable SSU transport (default = false)
# ssu = false
```

Related changes to documentation will be added if above changes will be accepted.


r4sas:
btw, why different ports? We can use both TCP and UDP on same port, isn't it?

polistern:
> btw, why different ports? We can use both TCP and UDP on same port, isn't it?

This change is more for explicitly specifying not only for TCP port, but also for UDP port.

For example, if it became necessary to open SAM outwards, even with address restrictions or authentication in future, an outsider might assume using I2Pd with SAM enabled.
Also, an explicit indication of both ports (for example, 53 UDP and 443 TCP) makes it possible to externally pass off the I2Pd as ordinary services.

I can add UPD port calculation from TCP port if UDP is not explicitly specified.
But then we need to decide: use the same TCP port for UDP or calculate it(like `TCPport - 1`)?

polistern:
Well, I think we can leave current UDP port calculation from TCP port (`TCPport - 1`) for the solid backwards compatible.

So if value for UDP port in config is `0` (by default), this tells us that the UDP port value will need to be calculated from TCP port (`TCPport - 1`).
If UDP port set explicitly in config - set from config.

I'll make these changes over this weekend.

PR #1797, opened 2022-10-30
Fix definitions for Darwin PPC; do not use pthread_setname_np on unsupported OS versions

barracuda156:
Darwin PPC has a define for both ppc and ppc64 which is `__POWERPC__`. Currently used one fails to apply and does nothing.

P. S. This is a necessary fix, however not sufficient. With correct defines in place, code in `util.cpp` uses a fallback, and build fails: https://github.com/PurpleI2P/i2pd/issues/1726#issuecomment-1296342335

@r4sas I will look for a suitable solution, we have this missing `pthread_*` problem recurring with multiple ports.

### DIFF
--- a/libi2pd/util.cpp
+++ b/libi2pd/util.cpp
@@ -22,6 +22,9 @@
 #include <pthread_np.h>
 #endif
 
+#if defined(__APPLE__)
+# include <AvailabilityMacros.h>
+#endif
 
 #ifdef _WIN32
 #include <stdlib.h>
@@ -143,8 +146,15 @@ namespace util
 	}
 
 	void SetThreadName (const char *name) {
-#if defined(__APPLE__) && !defined(__POWERPC__)
+#if defined(__APPLE__)
+# if (!defined(MAC_OS_X_VERSION_10_6) || \
+		(MAC_OS_X_VERSION_MAX_ALLOWED < 1060) || \
+		defined(__POWERPC__))
+		/* pthread_setname_np is not there on <10.6 and all PPC.
+		So do nothing. */
+# else
 		pthread_setname_np((char*)name);
+# endif
 #elif defined(__FreeBSD__) || defined(__OpenBSD__)
 		pthread_set_name_np(pthread_self(), name);
 #elif defined(__NetBSD__)
--- a/libi2pd/util.cpp
+++ b/libi2pd/util.cpp
@@ -143,7 +143,7 @@ namespace util
 	}
 
 	void SetThreadName (const char *name) {
-#if defined(__APPLE__) && !defined(__powerpc__)
+#if defined(__APPLE__) && !defined(__POWERPC__)
 		pthread_setname_np((char*)name);
 #elif defined(__FreeBSD__) || defined(__OpenBSD__)
 		pthread_set_name_np(pthread_self(), name);
--- a/build/cmake_modules/TargetArch.cmake
+++ b/build/cmake_modules/TargetArch.cmake
@@ -61,7 +61,7 @@ set(archdetect_c_code "
     #else
         #error cmake_ARCH mips
     #endif
-#elif defined(__ppc__) || defined(__ppc) || defined(__powerpc__) \\
+#elif defined(__ppc__) || defined(__ppc) || defined(__powerpc__) || defined(__POWERPC__) \\
       || defined(_ARCH_COM) || defined(_ARCH_PWR) || defined(_ARCH_PPC)  \\
       || defined(_M_MPPC) || defined(_M_PPC)
     #if defined(__ppc64__) || defined(__powerpc64__) || defined(__64BIT__)


libi2pd/util.cpp
```diff
@@ -143,7 +143,7 @@ namespace util
 	}
 
 	void SetThreadName (const char *name) {
-#if defined(__APPLE__) && !defined(__powerpc__)
+#if defined(__APPLE__) && !defined(__POWERPC__)
```
r4sas, line 146: Let's do this with lower-case and upper-case defines (3 checks).

r4sas (changes requested):

libi2pd/util.cpp
r4sas: Or maybe move out those and do nothing inside separate `#else if`...

barracuda156: @r4sas Since this is Apple-only condition, specifying that Linux definition is not used is redundant. (It won’t break anything, but it will do nothing.)
Apple used `__ppc__`, `__ppc64__` and `__POWERPC__`. All others are AIX, BeOS, Linux, *BSD etc.

r4sas: That's why I proposing to add empty block, like:
```cpp
#if defined(__APPLE__) && !defined(__powerpc__) && !defined(__POWERPC__)
		pthread_setname_np((char*)name);
#if defined(__APPLE__) && (defined(__powerpc__) || defined(__POWERPC__))
		// do nothing
#elif defined(__FreeBSD__) || defined(__OpenBSD__)
...
```

barracuda156: > Or maybe move out those and do nothing inside separate `#else if`...

@r4sas Ideally we gonna need something like was done in Ruby, to accommodate 10.5 Intel, for example, where that function is absent as well.

Ref: https://github.com/ruby/ruby/blob/master/thread_pthread.c#L1854

r4sas: > It won’t break anything, but it will do nothing.

Ah, yes. There no `#else` block. In that case only `!defined(__powerpc__) && !defined(__POWERPC__)` will be enough.

barracuda156: > That's why I proposing to add empty block, like:
> 
> ```c++
> #if defined(__APPLE__) && !defined(__powerpc__) && !defined(__POWERPC__)
> 		pthread_setname_np((char*)name);
> #if defined(__APPLE__) && (defined(__powerpc__) || defined(__POWERPC__))
> 		// do nothing
> #elif defined(__FreeBSD__) || defined(__OpenBSD__)
> ...
> ```

@r4sas Yeah, if do nothing is an acceptable behavior, this is certainly better than a broken fallback or using a non-existing function.

P. S. Do you have some reasons to use `__powerpc__` within Apple defines? I am not gonna insist here, but it is not needed. _Trust me, I know which macros work on Darwin PPC_ :)

barracuda156: > > It won’t break anything, but it will do nothing.
> 
> Ah, yes. There no `#else` block. In that case changing to `!defined(__powerpc__) && !defined(__POWERPC__)` will be enough.

There is, and it breaks the build:
```
#elif !defined(__gnu_hurd__)
		pthread_setname_np(pthread_self(), name);
#endif
```
Sure enough, `__gnu_hurd__` is not defined on Apple, whether PowerPC or not, so this fallback gets triggered.

r4sas: > Sure enough, `__gnu_hurd__` is not defined on Apple, whether PowerPC or not, so this fallback gets triggered.

Oops, didn't notice the negative (`!`) sign :D

> @r4sas Yeah, if do nothing is an acceptable behavior, this is certainly better than a broken fallback or using a non-existing function.

In that case lets do it this way. Thread names are not so important.

barracuda156: @r4sas Considering what Iain says here https://github.com/PurpleI2P/i2pd/issues/1726#issuecomment-1296353361 and that on 10.6 PPC (and 10.6.8 Rosetta) `pthread_setname_np` did not cause a build error, it will be better to make a do nothing fallback not for PPC, but rather for macOS < 10.6.

This will also fix the problem for old Intel systems, which likewise have no `pthread_setname_np`.

What do you think?

barracuda156:
@r4sas On a side note (I am not gonna deal with this in the PR), it appears that negative define is prone to failure. Now we got it on Darwin, all of a sudden, later it may break the build for someone else.

libi2pd/util.cpp
barracuda156: @r4sas As you suggested, I added a block for do-nothing-on-old-Darwin case, and then anything except those gonna have `pthread_setname_np`. A bit simplified logic from Ruby.

libi2pd/util.cpp
```diff
@@ -143,8 +146,15 @@ namespace util
 	}
 
 	void SetThreadName (const char *name) {
-#if defined(__APPLE__) && !defined(__powerpc__)
+#if defined(__APPLE__)
+# if (!defined(MAC_OS_X_VERSION_10_6) || \
```
r4sas, line 150: `not 10.6`??? It will stop checking here leading to part where is nothing is done.

barracuda156: @r4sas `MAC_OS_X_VERSION_10_6` _is_ defined for all macOS starting from 10.6. There is no forward definition. As you can see, on 10.5 it is _not_ defined: https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.5.sdk/usr/include/AvailabilityMacros.h

So the logic reads: if not defined `MAC_OS_X_VERSION_10_6` = all macOS prior to 10.6 (Intel has to be included, so having only `__POWERPC__` is insufficient).

PR #1816, opened 2022-12-02
Remove "X-Requested-With"

jiigen:
When any app uses Android’s WebView to load a web page, WebView attaches an extra header, named X-Requested-With, with the value set to the application ID.  That include Lightning-I2P browser by R4SAS and any browser based on Android Webview. 
Google doesn’t want to make it easy to get rid of the X-Requested-With header. However, there is a mechanism for replacing header information but anyway doesn’t allow a program to stop sending the X-Requested-With header.

More info on: https://www.stoutner.com/the-x-requested-with-header/

### DIFF
--- a/libi2pd_client/HTTPProxy.cpp
+++ b/libi2pd_client/HTTPProxy.cpp
@@ -252,6 +252,7 @@ namespace proxy {
 		req.RemoveHeader("From");
 		req.RemoveHeader("Forwarded");
 		req.RemoveHeader("DNT"); // Useless DoNotTrack flag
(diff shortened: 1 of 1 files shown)


r4sas (approved):

r4sas:
@jiigen I am thinking of changing this code, because it breaks web applications. PrivateBin is an [example of this](https://github.com/PrivateBin/PrivateBin/blob/master/lib/Request.php#L267-L268).

jiigen:
It seems to break ajax requests. I think that using javascript in privacy browsing is a very bad idea. However this bad habit to use web apps inside browser is very spread among users so I understand the reason.

PR #1828, opened 2022-12-24
Fix CheckAtomic.cmake

hack3ric:
This replaces the outdated LLVM CheckAtomic.cmake with [the new one](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CheckAtomic.cmake). Old CheckAtomic had false positives on RISC-V, where atomics of smaller sizes does not exist while larger ones do, which is fixed in a [commit](https://github.com/llvm/llvm-project/commit/cef85193b2cc1817ca43199a0ae9c6f25723997d).

This fixes build on riscv64 when testing on my QEMU machine.


r4sas:
Thatnk you!

PR #1901, opened 2023-03-05
Makefile.osx: do not use Intel flags on PPC

barracuda156:
Let’s finally unbreak this too :)

### DIFF
--- a/Makefile.osx
+++ b/Makefile.osx
@@ -23,9 +23,12 @@ ifeq ($(USE_UPNP),yes)
 	endif
 endif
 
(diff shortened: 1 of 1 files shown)


r4sas:
Don't you use cmake on PPC?

barracuda156:
> Don't you use cmake on PPC?

1. Yeah, but why have it broken for someone who uses Makefiles?
2. I bumped into this issue when trying to build `i2pd-tools` which have no CMake support.

barracuda156:
@r4sas Thank you!

PR #1909, opened 2023-03-31
Add critical log level

wekoq:
Add critical log level


orignal:
Most of errors are not critical

PR #1923, opened 2023-05-03
Convert raw x86_64 assembly to intrinsics, e2k hardware acceleration support

contextswap:
After talking with @makise-homura I've found out that Elbrus-8SV does in fact have hardware acceleration for AES-NI and AVX. As a result I have ported the raw x86_64 assembly to Intel intrinsics as Elbrus supports it. I have tested the code and it seems like I am not getting any checksum errors or packet corruptions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ ifneq (, $(DESTDIR))
 	PREFIX = $(DESTDIR)
 endif
 
+ifneq (, $(findstring e2k, $(SYS)))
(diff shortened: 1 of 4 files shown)


contextswap:
It seems like a few things are haywire in the AES. I'll fix and un-draft this again.

contextswap:
I finally fixed all the crypto issues. I think this can be merged now.

orignal:
don't reinnvent a wheel, use openssl's AESNI

PR #1931, opened 2023-05-15
Singlefile profiles db

hagen-i2p:
Patch to store peer profiles in one text file as json-hash-per-line

- Greatly saves disk space, reduces inode usage and fs writes (from 880Mb dir with ~210k 4k-files -> one file about 20Mb in my case)
- No extra deps: uses same boost/property_tree/json_parser as in I2PControl service
- Atomic database save: uses save-then-replace method

Drawbacks:

- In theory it uses slightly more memory just after start (as all known profiles loaded at startup). From the other side, profiles considered "useless" or "expired" will not be saved and live only in memory.
- No lazy save/load for individual profile, all profiles saved at once with PEER_PROFILE_AUTOCLEAN_TIMEOUT interval (6h for now, may be tuned later)

### DIFF
--- a/libi2pd/FS.cpp
+++ b/libi2pd/FS.cpp
@@ -270,6 +270,15 @@ namespace fs {
 		boost::filesystem::remove(path);
 	}
 
+	void HashedStorage::Destroy() {
+		std::string storage_root = GetRoot ();
+		if (storage_root.empty()) {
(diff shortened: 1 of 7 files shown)


wekoq:
I think, IsUseful can be inline.

PR #1949, opened 2023-07-11
Added undefined behavior sanitizer and fuzzers.

chadf:


fuzzing/fuzz-NetDb-HandleNTCP2RouterInfoMsg.cc
```diff
@@ -0,0 +1,30 @@
...
+{
+	i2p::I2NPMessageType msgType;
+
+
+	if(size < 1)
+		return true;
+
+	msgType = (i2p::I2NPMessageType) data[0];
+
+	data++;
```
brunoerg, line 23: Why this `data++`?

chadf: It skips over the message type value it took from the fuzzing input.

chadf:
Updated branch to be mergeable again.
Minimized changes to basic fuzzer only, to start.
Updated naming to be used easier in OSS-Fuzz (https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/).

PR #1954, opened 2023-07-21
Remove deprecated bind_at_load from macOS

rex4539:
Fix `ld: warning: -bind_at_load is deprecated on macOS`


r4sas:
How will this change affect the static builds for MacOS that are made with every release?

PR #1959, opened 2023-08-10
CPU: remove AVX code, switch to __builtin for AES detection

r4sas:
Removing outdated AVX support checks, one and only AVX asm block and switching to __builtin_cpu_supports to detect AES support.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -34,3 +34,6 @@ trim_trailing_whitespace = false
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.patch]
+trim_trailing_whitespace = false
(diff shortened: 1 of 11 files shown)


r4sas:
@khumarahn @yhaenggi @barracuda156
Can you check changes on your systems and platforms?

barracuda156:
@r4sas If we are to believe GCC documentation, these builtins are Intel-only: https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html
So I would expect this to fail. However I will check to make sure and update you.

r4sas:
Yes, I know that it is x86-only, but, as you can see, some architectures supports them too.
Anyway, I think it is better to limit checks to supported-only archs. I'll push some more changes now.

barracuda156:
@pkubaj If you got a minute, could you please comment on whether this works on the new POWER systems?

barracuda156:
From 7a8aa37eb2c45a96ced900d11103cfa0624cee24 it builds fine for me.

Issue #1965, opened 2023-08-31
IPv6 disabled by default


juliaszone:
Hi,
I've found IPv6 to be disabled by default in I2Pd which is a shame in my eyes. IPv6 is very important for the growth of the internet and I2P network.
We're approaching the point where IPv6 is becoming more and more widespread, so I believe ipv6=true should be the default.

Unfortunately, I don't know how I2Pd will behave on clients that do not have IPv6 connectivity, so this might be a good idea to look into. If it might break connectivity for those legacy clients it might be good to wait for a patch. I might be able to do some testing here later.

You can see below where it is set to false, meaning only Legacy IP will be used by default:
https://github.com/PurpleI2P/i2pd/blob/7b6aa41ca8dc3ca2a0c9d01ff9870794ad98e403/contrib/i2pd.conf#L77C11-L77C12
https://i2pd.readthedocs.io/en/latest/user-guide/configuration/#general-options

r4sas:
> so I believe ipv6=true should be the default

No. Less than 10% of users still have IPv6 with their ISPs.

wipedlifepotato:
IPv6 is trash

juliaszone:
> > so I believe ipv6=true should be the default
> 
> No. Less than 10% of users still have IPv6 with their ISPs.

Google reported 45.28% adoption globally on clients, either way, client adoption should not be a factor.
Clients that already have native connectivity should be able to use it out of the box. Only this really drives use, not many will go through config files.

juliaszone:
> IPv6 is trash

This is not useful in any way.

wipedlifepotato:
https://www.google.com/intl/en/ipv6/statistics.html
Ok there is a better way to see where is ipv6
чт, 7 сент. 2023 г., 04:09 Julia Logan ***@***.***>:

wipedlifepotato:
In some countries is really popular maybe there is not a trash. Yet much
countries not a have ipv6. If we want to check exists ipv6 or no we can to
create option "auto" and set up this option by default

чт, 7 сент. 2023 г., 10:02 Aleksandr Aleksandr ***@***.***>:

wipedlifepotato:
Not have an ipv6*/don't have ipv6/... *

чт, 7 сент. 2023 г., 10:05 Aleksandr Aleksandr ***@***.***>:

wipedlifepotato:
So in my country like even dont know what is ipv6 what is technology so for
my country is a trash and dont will be yes by default

[...]

PR #1966, opened 2023-09-07
IPv6 and IPv4 auto mode #1965 issue

wipedlifepotato:


wipedlifepotato:
нужно подключение к ипв6 1 раз делать в методе
для проверки брать пиры из netdb 
часть кода можно удалить

wipedlifepotato:
Thread 1 "i2pd" received signal SIGSEGV, Segmentation fault.
0x00000000004d39f0 in i2p::data::RouterInfo::IsNTCP2V6 (this=0x0) at libi2pd/RouterInfo.h:230
230				bool IsNTCP2V6 () const { return m_SupportedTransports & eNTCP2V6; };
(gdb) bt
#0  0x00000000004d39f0 in i2p::data::RouterInfo::IsNTCP2V6 (this=0x0) at libi2pd/RouterInfo.h:230
#1  0x00000000004cda26 in operator() (__closure=0x7fffffffd34c) at libi2pd/Config.cpp:539
#2  0x00000000004cdb69 in i2p::config::IsIPv6Works () at libi2pd/Config.cpp:545
#3  0x00000000005a7649 in std::__invoke_impl<bool, bool (*&)()> (__f=@0x7fffffffd4f0: 0x4cdb30 <i2p::config::IsIPv6Works()>) at /usr/include/c++/13/bits/invoke.h:61

wipedlifepotato:
Need a start netdb firstly but there is an another trouble

wipedlifepotato:
https://github.com/PurpleI2P/i2pd/commit/d0316f5bdd1b1c15b16e2eb863c856e42ed2b12a#diff-e50973d72af06df6214f78b71992c591d5fce3a8198d85ba58713742717d71eeR538
возвращает нуллевой указатель нужно сначала инициализацию сделать но проблема в том что тогда будет циклично вызывать метод нужно обдумать как лучше сделать и гугл запрашивать тоже плохая идея просто к ::1 цеплятся тоже плохая идея проверять по интерфейсам тоже плохая идея

wipedlifepotato:
15:32:26@838/debug - Transports: IPv6 supports is 0
15:32:26@838/debug - Transports: IPv4 supports is 0

If do connection to 0.0.0.0/127.0.0.1 then is a bad idea. So only get from netdb maybe run this tests in netdb code need a think

wipedlifepotato:
https://github.com/PurpleI2P/i2pd/blob/openssl/libi2pd/Transports.cpp#L669

wipedlifepotato:
bool IsIPv4Works(void)
	{
	       if(! i2p::context.GetTesting () ) return true; // For a test
	       auto retVal = i2p::context.GetStatus();
               LogPrint (eLogDebug, "Transports: IPv4 supports is ", retVal);
	       return retVal;
	}
	bool IsIPv6Works(void)
	{
	       if(! i2p::context.GetTestingV6 () ) return true; // For a test
	       auto retVal= i2p::context.GetStatusV6();
               LogPrint (eLogDebug, "Transports: IPv6 supports is ", retVal);
	       return retVal;
	}


If we do simillar code, then is like to set up "true" always for ipv6

juliaszone:
I've tried your latest commit and `./i2pd --loglevel=debug` is stuck at no output, with `ss -tupn` showing
`tcp         SYN-SENT       0            1                         [2a02:810d:b5bf:ece1:5054:ff:fea4:e2b8]:51512                  [2001:4860:4860::8844]:*           users:(("i2pd",pid=10188,fd=6))               `.
The host I'm trying this on has working IPv6.
I have been checking the tcpdump and it looks like you're trying to connect to Google DNS on port 0. I can try to troubleshoot this and let you know if I got a fix.

juliaszone:
We can also try to make an http request for ipv4.icanhazip.com and ipv6.icanhazip.com, as that's backed by cloudflare now as it seems.

wipedlifepotato:
Ok. If you can, you can git clone my repository and pull requests if you
dont know how to do it. then it is ok. I can too. Firstly I would like to
ask original and r4sas in iRc about our idea. What did you say?

пт, 8 сент. 2023 г. в 19:14, Julia Logan ***@***.***>:

arthurmelton:
I would say that we should try and not rely on some third-party service to be up. I feel like the best solution would be to check the device interfaces and see if we have an IPv4/6 IP address. Someone could have the Google IPv6 DNS servers blocked, but still have IPv6 on the machine.

PR #2037, opened 2024-03-08
fix HashedStorage::Init exceptions in ios simulator.

vovasty:
On iOS simulator `HashedStorage::Init` throws an exception `file exists` cause FS is case insensitive.
In order to fix this we use non-throwing variant of `boost::filesystem::create_directory` and ignore `boost::system::errc::file_exists` error code.

### DIFF
--- a/libi2pd/FS.cpp
+++ b/libi2pd/FS.cpp
@@ -256,7 +256,7 @@ namespace fs {
 			auto p = root + i2p::fs::dirSep + prefix1 + chars[i];
 			if (boost::filesystem::exists(p))
 				continue;
-#ifdef TARGET_OS_SIMULATOR
+#if TARGET_OS_SIMULATOR
             // ios simulator fs says it is case sensitive, but it is not
             boost::system::error_code ec;
             if (boost::filesystem::create_directory(p, ec))
--- a/libi2pd/FS.cpp
+++ b/libi2pd/FS.cpp
@@ -9,6 +9,11 @@
 #include <algorithm>
 #include <boost/filesystem.hpp>
 
+#if defined(MAC_OSX)
+#include <boost/system/system_error.hpp>
+#include <TargetConditionals.h>
+#endif
+
 #ifdef _WIN32
 #include <shlobj.h>
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


Vort:
So what happens? OS changes letter cases when directory is created, but then can't find directory because it became case sensitive?
I though sensitivity is related only to name matching, not to changing file case during creation.

vovasty:
OS doesn't change the case. When you create `FolderName` it is `FolderName`, but if you try to create `foldername` it fails cause folder exists.
In other words `boost::filesystem::exists` falsibly says that folder doesn't exists but `boost::filesystem::create_directory` throws an exception.

Vort:
> When you create FolderName it is FolderName, but if you try to create foldername it fails cause folder exists.

I wonder why such situation may happen in i2pd.

vovasty:
> I wonder why such situation may happen in i2pd.

on i2pd start it calls `HashedStorage::Init` which creates folders `dataDir/netDb` in form `r` + letters from i2p::data::GetBase64SubstitutionTable() which are upper/lower case.

libi2pd/FS.cpp
```diff
@@ -251,8 +256,22 @@ namespace fs {
 			auto p = root + i2p::fs::dirSep + prefix1 + chars[i];
 			if (boost::filesystem::exists(p))
 				continue;
-			if (boost::filesystem::create_directory(p))
+#ifdef TARGET_OS_SIMULATOR
```
Vort, line 259: From what I found, in `TargetConditionals.h` `TARGET_OS_SIMULATOR` can be defined as `0`, which counts as _defined_ for `#ifdef`. I think `#if` should be used here instead. @vovasty please recheck.
https://github.com/xybp888/iOS-SDKs/blob/6d733d8ec34d59a9b13d8e374b199bd3a6630a39/iPhoneOS13.0.sdk/usr/include/TargetConditionals.h#L143

libi2pd/FS.cpp
```diff
@@ -9,6 +9,11 @@
 #include <algorithm>
 #include <boost/filesystem.hpp>
 
+#if defined(MAC_OSX)
```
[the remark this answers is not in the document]
Vort: Looks like `TargetConditionals.h` is the place where `TARGET_OS_SIMULATOR` value should be stored. And such file is not available on other OSes, so this part of the code looks ok.

libi2pd/FS.cpp
vovasty: Fixed. Thanks of spotting this!

PR #2053, opened 2024-04-12
change exec path to /usr/bin in i2pd.service

yangfl:
CMakeLists.txt already installs i2pd to CMAKE_INSTALL_BINDIR, which is normally /usr/bin.

Change i2pd.service to match the behavior of cmake build script.


r4sas:
https://github.com/PurpleI2P/i2pd/discussions/2068

PR #2101, opened 2024-09-25
Added example docker-compose.yml

imdef:
I added an example docker-compose.yml file that I used in my own setup to persist the router's data because I read "You must save the router's data (netdb, configuration, etc.) between runs of the router. I2P does not work well if you must reseed each startup, and that's a huge load on our reseed servers, and not very good for anonymity either. Even if you bundle router infos, I2P needs saved profile data for best performance. Without persistence, your users will have a poor startup experience."

### DIFF
--- a/contrib/docker/docker-compose.yml
+++ b/contrib/docker/docker-compose.yml
@@ -1,0 +1,13 @@
+services:
+  i2pd:
+    container_name: i2pd2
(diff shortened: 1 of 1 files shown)


Raimo33:
This is off topic and not worth including. Also, the docker-compose configuration is subjective

PR #2140, opened 2024-12-26
systemd: Add yggdrasil.service as a weak dependency to i2pd.service

lcharles123:


mittwerk:
does that change anything or maybe fix something?

Darkcyankitty:
> does that change anything or maybe fix something?

Useful when Yggdrasil only mode or just used yggdrasil network with i2pd.

r4sas:
I think we don't need that.

PR #2169, opened 2025-03-31
Fix problem with i2pd service

ChernovHere:
My case: 

When changing the address of the web interface or http/socks proxy to `192.168.X.X` and after restarting the system, an error pops up: 

```
XX:XX:XX:XX@365/critical - Daemon: Failed to start Webconsole: bind: Cannot assign requested address
XX:XX:XX:XX@365/critical - Clients: Exception in HTTP Proxy: bind: Cannot assign requested address
XX:XX:XX:XX@365/critical - Clients: Exception in SOCKS Proxy: bind: Cannot assign requested address
```

This PR fixes the problem

### DIFF
--- a/contrib/i2pd.service
+++ b/contrib/i2pd.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=I2P Router written in C++
 Documentation=man:i2pd(1) https://i2pd.readthedocs.io/en/latest/
-After=network.target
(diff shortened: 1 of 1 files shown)


r4sas:
Может в Wants оставить `network.target`?

```
Wants=network.target
After=network.target network-online.target
```

PR #2211, opened 2025-07-14
Update deprecated openssl functions and remove warning suppression

Houkime:
see https://docs.openssl.org/3.0/man7/migration_guide/


orignal:
I was in the middle of this change anyway. But fine.

PR #2245, opened 2025-10-04
Mycelium network support

oooo-ps:
This is just the initial commit that implements the Mycelium network condition, which is quite similar to Yggdrasil but operates in the `0400::/7` range:

https://github.com/threefoldtech/mycelium

I have a few questions:

1. does it make sense to make this contribution?
2. is the `openssl` branch the correct one to fork the upstream?

UPD. While working on the second part, I found only one issue related to supporting multiple mesh networks: the router uses a single registry called `eNTCP2V6MeshIdx`. It might make sense to rename Yggdrasil members to "Mesh" or to separate the indexes into different stacks. Configuration options would depend on this implementation, as there are too many potential networks possible to make the router extensible without increasing the codebase.

### DIFF
--- a/libi2pd/util.cpp
+++ b/libi2pd/util.cpp
@@ -636,6 +636,17 @@ namespace net
 		return IsYggdrasilAddress (addr.to_v6 ().to_bytes ().data ());
 	}
 
+	static bool IsMyceliumAddress (const uint8_t addr[16])
+	{
+		return addr[0] == 0x04 || addr[0] == 0x05;
+	}
+
+	bool IsMyceliumAddress (const boost::asio::ip::address& addr)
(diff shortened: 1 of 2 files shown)


orignal:
1. No worth to do it at this point until it reaches few thousands routers
2. Yes, openssl is right branch

i2pd was initially designed to support multiple mesh network Yes you need to rewrite it in many places, that's why it's not worth to do. A lot of work with tiny or no result.

orignal:
>but I don't know how to collect the DB if my router running Yggdrasil-only mode.
No difference

But I meant to implement the code in Reseed.cpp to support mycelium reseeds if you have a local interface with it.

PR #2265, opened 2025-11-23
contrib/openrc: improve `command_args`

fkobi:
It looks better now!


r4sas:
Nope, better leave `mainconf`, `tunconf`, `tundir` as is, because some distributions can re-write that values with awk/grep (like Entware, OpenWRT).

fkobi:
I will leave them as-is but open another PR as I think there still room for improvement there.

Issue #2243, opened 2025-10-02
NAT type gets detected wrong


HardenedSteel:
I tried I2Pd on both full-cone NAT and port restricted cone NAT but I2Pd always reports NAT type as symmetric NAT.

orignal:
Then please tell how it should be done properly.
https://github.com/PurpleI2P/i2pd/blob/openssl/libi2pd/SSU2Session.cpp#L1832-L1841

HardenedSteel:
The current logic only seems to compare the result from **a single peer**. The logic should compare the port number from the results obtained from **multiple peers** to check if it remains the same or changes.

With this we can detect if the NAT is cone (when same) or symmetric (when changes).

A tool for detecting NAT type: https://github.com/pion/stun/tree/master/cmd/stun-nat-behaviour

PR #2272, opened 2025-12-13
feat: change published port if full-cone

HardenedSteel:
If we're behind full-cone NAT; update our published port number while keeping our listening port number same, so other peers can connect us when we're behind NAT.

Depends on: #2243 #2273

### DIFF
--- a/libi2pd/SSU2Session.cpp
+++ b/libi2pd/SSU2Session.cpp
@@ -2444,7 +2444,9 @@ namespace transport
 												session->SendPeerTest (6, buf + offset, len - offset, addr, true);
 											}	
 											SetTestingState (false);
-											if (GetRouterStatus () != eRouterStatusFirewalled && addr->IsPeerTesting ())
+											if (i2p::context.GetError () == eRouterErrorFullConeNAT)
+												SetRouterStatus (eRouterStatusOK);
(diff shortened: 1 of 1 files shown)


libi2pd/SSU2Session.cpp
```diff
@@ -1831,20 +1831,23 @@ namespace transport
...
-					LogPrint (eLogInfo, "SSU2: Our port ", ep.port (), " received from ", m_RemoteEndpoint, " is different from ", m_Server.GetPort (isV4));
+					// TODO: External port detected incorrectly.
+					// LogPrint (eLogInfo, "SSU2: Our port ", ep.port (), " received from ", m_RemoteEndpoint, " is different from ", m_Server.GetPort (isV4));
 					if (isV4)
 					{
 						if (i2p::context.GetTesting ())
 							i2p::context.SetError (eRouterErrorSymmetricNAT);
 						else if (m_State == eSSU2SessionStatePeerTest)
 							i2p::context.SetError (eRouterErrorFullConeNAT);
+							i2p::context.UpdatePort((int)ep.port ());
```
orignal, line 1843: Brackets?  It's not Python. If assume it must belong to if (m_State == eSSU2SessionStatePeerTest) block.

PR #2308, opened 2026-01-29
Update Hebrew localization strings

sjehuda:
* Correct an error.
* Correct time sequence.
* Improve context.


orignal:
It should be done here https://crowdin.com/project/i2pd

sjehuda:
Done.

I hope that it was saved.

sjehuda:
The system of crowdin does not allow me to change the order of the
singular forms of
```
%d second
%d minute
%d hour
%d day
```

orignal:
Wait when @r4sas has time to update the localization

r4sas:
@sjehuda which order must be here? Like `second %d` (in Hebrew)?

r4sas:
I now approved you translation on Crowdin, please check it.

sjehuda:
> @sjehuda which order must be here? Like `second %d` (in Hebrew)?

@r4sas the order for singular should be reversed `%d second`.

As with the languages Arabic and Farsi, Hebrew is also a semitic
language which is written from right to left.

So it is to reverse the order for a reversed language.

I will try again at Crowdin; and, if it fails, again; then, I will send
the strings, here, for you to act upon it.

Thank you.

PR #2324, opened 2026-02-11
contrib/openrc/i2pd.openrc: remove `name`

fkobi:
That variable is no longer used:
https://github.com/OpenRC/openrc/blob/master/service-script-guide.md

### DIFF
--- a/contrib/openrc/i2pd.openrc
+++ b/contrib/openrc/i2pd.openrc
@@ -6,7 +6,6 @@ mainconf="/etc/i2pd/i2pd.conf"
 tunconf="/etc/i2pd/tunnels.conf"
 tundir="/etc/i2pd/tunnels.conf.d"
 
-name="i2pd"
 command="/usr/bin/i2pd"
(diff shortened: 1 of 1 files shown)


r4sas:
Have you tested it with whole bunch of targeted systems? OpenWRT and Entware included.

fkobi:
no, I have not.
OpenWRT does not use OpenRC though:
https://openwrt.org/docs/techref/initscripts#example_init_script
Note the shebang.

Also I have asked the devs on IRC if this is needed:
```irc
<fkobi> when is the variable `name` used in the init scripts?
<navi> the default `start` and `stop` implementations for start-stop-daemon
       and supervise-daemon use it to print out `Starting $name`
<navi> if `name` is unset, it falls back to `RC_SVCNAME`
<fkobi> so if $name = $RC_SRVNAME there is no point in defining it?
<navi> correct
```
Here `$name = $RC_SRVNAME` so it should be removed.

PR #2347, opened 2026-03-06
Create GPL-2.0.txt

bkmgit:
License for manual page.


orignal:
Our license is BSD. Why GPL?

bkmgit:
Man page is GPL:
https://github.com/PurpleI2P/i2pd/blob/openssl/debian/i2pd.1#L128

PR #2349, opened 2026-03-10
[debian] exclude obsolete dependencies

LLE8:
As far as I understand, support for Debian 10 was dropped because of the end of gcc 8 support, unless install gcc 10 in Buster from debian-backports (if possible) 
https://distrowatch.com/table.php?distribution=debian
So some packages are not required.
See also https://github.com/PurpleI2P/i2pd/issues/2262

Not tested on Debian 11.


LLE8:
Документацию чуть поправил.
Русская версия похоже очень старая?

PR #2366, opened 2026-04-04
pledge and unveil on OpenBSD

amrfti:
This patch implements native OpenBSD hardening for the i2pd daemon with [`unveil(2)`](https://man.openbsd.org/unveil.2) and [`pledge(2)`](https://man.openbsd.org/pledge.2) by placing a privledge restriction block at the end of `Daemon::init`. We get paths for `tunconf`, `tunnelsdir`, `datadir`, and `certsdir` from the config map. 

Paths for name resolution (`/etc/resolv.conf` and `/etc/hosts`) are [automatically unveiled](https://man.openbsd.org/pledge.2#dns~2) by the `dns` promise.

### DIFF
--- a/libi2pd/Destination.cpp
+++ b/libi2pd/Destination.cpp
@@ -161,10 +161,12 @@ namespace client
 		CleanUp (); // GarlicDestination
 	}
 
-	bool LeaseSetDestination::Reconfigure(const i2p::util::Mapping& params)
-	{
-		params.Get (I2CP_PARAM_DONT_PUBLISH_LEASESET, m_IsPublic);
 
+	bool LeaseSetDestination::Reconfigure (const i2p::util::Mapping& params)
+	{
+		bool dontPublishLeaseSet = !m_IsPublic;
+		params.Get(I2CP_PARAM_DONT_PUBLISH_LEASESET, dontPublishLeaseSet);
+		m_IsPublic = !dontPublishLeaseSet;
 		auto numTags = GetNumTags ();
 		params.Get (I2CP_PARAM_TAGS_TO_SEND, numTags);
 		auto numRatchetInboundTags = GetNumRatchetInboundTags ();


amrfti:
Upon further testing I believe this still has some issues. In my first commit, I was lacking some additional paths that may have conditionally needed to be unveiled based on given arguments. Yet even still there are some strange issues that I believe are a result of some arguments not being initialized to their default fallback options until *after* we've finished unveiling. For example, I created this error while testing:
```
# i2pd --logfile /root/.i2pd/i2pd.log
i2pd: Could not create pid file /root/.i2pd/i2pd.pid: No such file or directory
```
I believe at the time I unveil the `pidfile` path, it's unintialized, which causes it to never get unveiled. I need to do more testing, feedback welcome.

orignal:
daemon/Daemon.cpp:317:76: error: function definition is not allowed here
  317 |                 static void UnveilConfigOption(const std::string& key, const char* mask) {
      |                                                                                          ^
daemon/Daemon.cpp:337:4: error: use of undeclared identifier 'UnveilConfigOption'
  337 |                         UnveilConfigOption(rule.first, rule.second);
      |                         ^

Please refrain from using nested functions.

sonya-blade:
suppose it'd be best to go the "official" way:
  - submit the patch to the <ports@openbsd.org>, where it'd be reviewed and tested;
  - after the acceptance of the patch by community - merge in upstream from OpenBSD cvsweb or git
 thank you

daviduhden (changes requested):
Thanks for working on OpenBSD hardening. I agree with the issue reported by orignal, and here is the precise correction.

C++ does not allow defining a named function inside another function. In this patch, static void UnveilConfigOption(...) is defined inside Daemon::init, which causes the compile error: function definition is not allowed here.

Please move that helper out of Daemon::init, for example to anonymous namespace file scope, or replace it with a local lambda and keep the rest of the logic unchanged.

Also please fix the PR text typo: privledge -> privilege.

PR #2375, opened 2026-04-26
Add support for SAM over UNIX domain sockets

### LINKED ISSUE #1751: Add support for SAM over UNIX domain sockets
One might not want to every user under a system to have access to the i2pd daemon. UNIX domain sockets work great in this scenario as you can use traditional user administration (users and groups) to limit access to sockets. For instance, I could create an "`i2p-users`" group and only allow users under this group to have access to the SAMv3 socket.

Lkzeu:
closes #1751

### DIFF
--- a/libi2pd_client/SAM.cpp
+++ b/libi2pd_client/SAM.cpp
@@ -74,6 +74,33 @@ namespace client
 		m_Owner.RemoveSocket(shared_from_this());
 	}
 
+	void SAMSocket::GenerateRemoteEndpointStr (boost::system::error_code& ec) {
+		std::ostringstream oss;
+		auto tmp = m_Socket.remote_endpoint (ec);
+		if (ec) return;
+
+		if (tmp.protocol().family() == AF_INET || tmp.protocol().family() == AF_INET6)
+		{
+			boost::asio::ip::tcp::endpoint ep;
+			ep.resize(tmp.size());
+			std::memcpy(ep.data(), tmp.data(), tmp.size());
+			oss << ep;
+		}
+#if defined(BOOST_ASIO_HAS_LOCAL_SOCKETS)
+		else
+		{
+			boost::asio::local::stream_protocol::endpoint ep;
+			ep.resize(tmp.size());
+			std::memcpy(ep.data(), tmp.data(), tmp.size());
+			if (ep.path().empty())
+				oss << "unix:/abstract";
+			else
+				oss << ep.path();
+		}
+#endif
+		m_RemoteEndpointStr = oss.str();
+	}
+
 	void SAMSocket::ReceiveHandshake ()
 	{
 		m_Socket.async_read_some (boost::asio::buffer(m_Buffer, SAM_SOCKET_BUFFER_SIZE),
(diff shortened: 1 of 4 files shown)


libi2pd_client/SAM.cpp
```diff
@@ -74,6 +74,33 @@ namespace client
...
+	void SAMSocket::GenerateRemoteEndpointStr (boost::system::error_code& ec) {
+		std::ostringstream oss;
+		auto tmp = m_Socket.remote_endpoint (ec);
+		if (ec) return;
+
+		if (tmp.protocol().family() == AF_INET || tmp.protocol().family() == AF_INET6)
+		{
+			boost::asio::ip::tcp::endpoint ep;
+			ep.resize(tmp.size());
+			std::memcpy(ep.data(), tmp.data(), tmp.size());
```
orignal, line 86: Why do you use memcpy instead just ep = tmp?

libi2pd_client/SAM.h
```diff
@@ -314,6 +318,73 @@ namespace client
 			// for HTTP
 			const decltype(m_Sessions)& GetSessions () const { return m_Sessions; };
 	};
+
+	class SAMAcceptor
```
orignal, line 322: Better to use  template ServiceAcceptor   from I2PService.h implementing acceptors for different socket types.

libi2pd_client/SAM.cpp
Lkzeu: > Why do you use memcpy instead just ep = tmp?

The `SAMSocket::Socket_t` is now a `generic::stream_protocol::socket` to allow it to receive a `tcp::socket` or `local::stream_protocol::socket` at runtime.
So `remote_endpoint()` now returns a generic endpoint, which could not be copied into `tcp::endpoint`.

libi2pd_client/SAM.h
Lkzeu: > Better to use template ServiceAcceptor from I2PService.h implementing acceptors for different socket types.

Understood. SAMBridge needs to select the type of m_Acceptor at runtime based on the config file like I2PControl.h/cpp does it.

By using the polymorphic SAMAcceptor, SAMBridge uses only one m_Acceptor in the code instead of m_Acceptor and m_LocalAcceptor  like I2PControlService does.

I could add UDSAcceptor that derives from ServiceAcceptor and adapt SAMBridge/SAMSocket accordingly. but, as a template, I will need to check if the acceptor is m_Acceptor or m_LocalAcceptor.

Or I could work on a major refactor to make ServiceAcceptor polymorphic, similar to how SAMAcceptor is now.

What do you think?

libi2pd_client/SAM.cpp
orignal: If generic_endpoint is the base class, why not use static_cast instead? Relying on internals is always a bad idea, because it can change later.

libi2pd_client/SAM.h
orignal: Polymophic acceptor makes sense. I would implement it in more generic code (I2PService), because it's need in many other places beside SAM. And SOCKS proxy is the first in the list.

PR #2389, opened 2026-05-24
contrib: add sysusers.d file and update tmpfiles.d config file

bluca:
sysusers.d/tmpfiles.d config files allow a package to use declarative configuration instead of manually written maintainer scripts.
This also allows image-based systems to be created with /usr/ only, and also allows for factory resetting a system and recreating /etc/ on boot.

https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html
https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html

They are not specific to debian, so move them to the contrib/ directory.
Also rename to the canonical form, which is tmpfiles (plural).


orignal:
@r4sas please review

PR #2401, opened 2026-06-08
OpenBSD hardening: implement unveil/pledge correctly in daemon init

daviduhden:
This PR implements OpenBSD hardening (`unveil(2)` + `pledge(2)`) in `Daemon::init`, based on the intent of #2366, with runtime-path handling aligned to i2pd behavior.

## What this changes

- Adds OpenBSD-only hardening in `daemon/Daemon.cpp` (guarded by `#ifdef __OpenBSD__`).
- Applies `unveil()` to effective i2pd runtime paths (resolved from config/defaults), including:
  - main config
  - datadir
  - certsdir
  - tunconf
  - tunnelsdir
  - pidfile
  - logfile
  - local reseed files
- Locks unveil rules with `unveil(nullptr, nullptr)`.
- Applies `pledge("stdio inet dns flock rpath wpath cpath", nullptr)`.
- Handles hardening failures with explicit logging and clean init failure (`return false`).

## Why this version

The implementation uses resolved runtime paths rather than only raw config keys, so defaults used by i2pd are covered consistently. It also avoids treating `https://` reseed URLs as filesystem paths for unveil rules.

### DIFF
--- a/daemon/Daemon.cpp
+++ b/daemon/Daemon.cpp
@@ -10,6 +10,7 @@
 #include <memory>
 #include <regex>
 #include <array>
+#include <utility>
 
 #include "Daemon.h"
 
@@ -355,8 +356,8 @@ namespace util
 			{reseedFile, "r"},
 			{reseedZipFile, "r"}
 		}};
-		for (const auto& [path, mask]: unveilRules)
-			unveilPath (path, mask);
+		for (const auto& rule: unveilRules)
+			unveilPath (rule.first, rule.second);
 
 		if (unveil (nullptr, nullptr) == -1)
 		{
--- a/daemon/Daemon.cpp
+++ b/daemon/Daemon.cpp
@@ -9,6 +9,7 @@
 #include <thread>
 #include <memory>
 #include <regex>
+#include <array>
 
 #include "Daemon.h"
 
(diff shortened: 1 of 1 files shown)


daviduhden:
Verified on OpenBSD 7.9 with a clean build, unit tests, and a real runtime trace under `ktrace`.

Build and tests:

```sh
cmake -S build -B build-openbsd -G "Unix Makefiles" \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DBUILD_TESTING=ON

gmake -C build-openbsd -j "$(sysctl -n hw.ncpuonline 2>/dev/null || sysctl -n hw.ncpu || echo 4)"

ctest --test-dir build-openbsd --output-on-failure
```

Result:

```text
13/13 tests passed
```

Runtime verification with a minimal config and kernel tracing:

* `ktrace` showed multiple `unveil(...)` calls from `i2pd`.
* `kdump` confirmed `unveil(NULL, NULL)` to lock the unveil table.
* `kdump` confirmed `pledge("stdio inet dns flock rpath wpath cpath", NULL)`.
* All `pledge(2)` and `unveil(2)` calls returned `0`.
* `kdump -t p` showed no pledge violations.
* The webconsole came up successfully on `127.0.0.1:7070`.

In short: the OpenBSD hardening reaches the kernel correctly and does not break normal startup.

PR #2411, opened 2026-06-17
Haiku OS gui

wipedlifepotato:
Добавлена кнопка для запуска веб интерфейса, заготовлен код для получения информации о туннеле

### DIFF
--- a/daemon/HaikuDaemon.cpp
+++ b/daemon/HaikuDaemon.cpp
@@ -34,7 +34,8 @@ enum
 	M_RUN_PEER_TEST,
 	M_DUMMY_COMMAND,
 	C_GRACEFUL_SHUTDOWN_UPDATE,
-	C_MAIN_VIEW_UPDATE
+	C_MAIN_VIEW_UPDATE,
+	M_SHOW_TUNNEL
 };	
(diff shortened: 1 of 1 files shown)


daemon/HaikuDaemon.cpp
```diff
@@ -104,14 +120,79 @@ void MainWindow::UpdateMainView ()
...
+	std::string address; i2p::config::GetOption("http.address", address);
+	if(!enabled)
+	{
+		return m_MainView->SetText("Not enabled http server");
+	}
+//	int pid;
+	std::ostringstream com;
+	com << "WebPositive http://" << address << ":" << port;
+	//char * argv[] = {(char*)url.str().c_str() , NULL, NULL};
+	//BRoster{}.Launch( "application/x-vnd.Haiku-WebPositive", 2, argv);
```
hfsfox, line 173: I think that can be launched with other signature or through "open" shell command, which also can work with links

wipedlifepotato:
Run system on "open" Program? This program in $PATH? I TRied roster, but is
not runs like I want u can see last pr with fix. Because before I copy from
VM corrupted file my bad
So, if u know how to run this thought roster u can help 🐚

чт, 18 июн. 2026 г., 02:02 hfsfox ***@***.***>:

wipedlifepotato:
It's not*

чт, 18 июн. 2026 г., 02:07 Aleksandr Aleksandr ***@***.***>:

PR #2423, opened 2026-06-27
feat: Add Clang/LLVM LTO optimization support with native architecture tuning

ETJAKEOC:
- Add LTO mode selection (thin/full) with thin as default for faster builds
- Add native architecture tuning (-march=native -mtune=native) support
- Clang-specific vectorization and optimization flags
- Improve debug builds to include -O2 with debug symbols
- Increase default -Os to -O3 for release builds
- Add convenient make targets for optimized builds (optimize-thin, optimize-full)
- Support native architectures via MARCH/MTUNE vars
- Maintain backward compatibility with existing build system

I have tested this build on a bdver2 system, and while the binary did seem to increase in size of the binary about 5X, the service still seems to run. Further testing by more familiar experts of this program should probably be done to ensure that everything is proper on other MARCH/MTUNE builds.


orignal:
we don't need this

PR #2426, opened 2026-06-30
Add transparent proxying

smurfix:
This patch series adds support for transparent proxying: a small DNS resolver translates .i2p domains to addresses in a special subnet, which gets rerouted (by the Linux kernel) to i2pd.

Thus you can set up the infrastructure in your router once, then every client can connect through i2p directly. No client configuration required.

Inspired by the equivalent feature in TOR. Authored by Claude 4.8 (plan) and GLM 5.2 (code, testing).


orignal:
AI slop is not allowed.

Beside it, this functionality is not needed, that was  discussed long time ago.

PR #2445, opened 2026-07-25
chore: fix issues found through static analysis

jpk68:
Fixes a number of small issues found using `scan-build` and `clang-tidy`.

### DIFF
--- a/libi2pd/Config.h
+++ b/libi2pd/Config.h
@@ -83,7 +83,10 @@ namespace config {
 	bool GetOption(const char *name, T& value)
 	{
 		if (!m_Options.count(name))
+		{
+			value = T();
 			return false;
+		}
 		value = m_Options[name].as<T>();
 		return true;
 	}
--- a/libi2pd/TunnelConfig.cpp
+++ b/libi2pd/TunnelConfig.cpp
@@ -178,7 +178,10 @@ namespace tunnel
 	{
 		// inbound
 		CreatePeers (peers);
-		m_LastHop->SetNextIdent (i2p::context.GetIdentHash ());
+		if (m_LastHop)
+			m_LastHop->SetNextIdent (i2p::context.GetIdentHash ());
+		else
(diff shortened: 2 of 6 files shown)


jpk68:
@orignal Also, please check if you have received my email.

orignal:
Better to fix 
template<typename T>
bool GetOption(const char *name, T& value)

instead initializing it before every call

orignal:
> please check if you have received my email.

Please don't waste my time with neuroslop.

jpk68:
> > please check if you have received my email.
> 
> Please don't waste my time with neuroslop.

Respectfully, you are mistaken. I think it's irresponsible to respond in such a way to legitimate security disclosures which could compromise the safety of your users.

About the actual issues themselves: multiple cases I gave are textbook-example off-by-one reads; a literal answer to your request for "real cases when it goes beyond [the buffer]." Saying that my email is AI-generated is not a valid rebuttal.

Since it looks like you won't be able to handle security disclosures responsibly, I will be forced to publicly submit the patches instead. Please try and do better.

wipedlifepotato:
For fix security things need an exploit for example.

orignal:
Regradless, if it's AI generated or nor, I don't waste time reading long texts, a priori assuming that this is a neuroslope.

wipedlifepotato:
What for there is changes for hops logic? What these analyzator said

jpk68:
> Regradless, if it's AI generated or nor, I don't waste time reading long texts, a priori assuming that this is a neuroslope.

This is just ridiculous and extremely poor practice.

orignal:
> This is just ridiculous and extremely poor practice.

May be. But I have too limited time.

jpk68:
> May be. But I have too limited time.

I understand. The only thing I ask is that you take it seriously and read through the email when you have time. These are serious bugs that need to be fixed; I don't think it would be good to fix them openly in a PR.

PR #2446, opened 2026-07-25
chore: fix memory leaks

jpk68:
Fixes two memory leaks found with `valgrind`.

### DIFF
--- a/libi2pd/Destination.cpp
+++ b/libi2pd/Destination.cpp
@@ -166,6 +166,7 @@ namespace client
 		m_CleanupTimer.cancel ();
 		m_PublishConfirmationTimer.cancel ();
 		m_PublishVerificationTimer.cancel ();
+		m_PublishDelayTimer.cancel ();
 		if (m_Pool)
 		{
 			m_Pool->SetLocalDestination (nullptr);
--- a/libi2pd/SSU2OutOfSession.cpp
+++ b/libi2pd/SSU2OutOfSession.cpp
@@ -170,6 +170,7 @@ namespace transport
 		header.h.flags[2] = 0; // flag
 		memcpy (h, header.buf, 16);
 		htobuf64 (h + 16, GetSourceConnID ()); // source id
+		memset (h + 24, 0, 8); // header token
 		// payload
 		payload[0] = eSSU2BlkDateTime;
 		htobe16buf (payload + 1, 4);


libi2pd/SSU2OutOfSession.cpp
```diff
@@ -170,6 +170,7 @@ namespace transport
 		header.h.flags[2] = 0; // flag
 		memcpy (h, header.buf, 16);
 		htobuf64 (h + 16, GetSourceConnID ()); // source id
+		RAND_bytes (h + 24, 8); // header token
```
orignal: Do we need token for peer test? Is memset (h + 24, 0, 8) not  enough?

libi2pd/Destination.cpp
```diff
@@ -166,6 +166,7 @@ namespace client
 		m_CleanupTimer.cancel ();
 		m_PublishConfirmationTimer.cancel ();
 		m_PublishVerificationTimer.cancel ();
+		m_PublishDelayTimer.cancel ();
```
orignal, line 169: Why does this fix memory leak?
In worst case time will terminate by timeout.

libi2pd/SSU2OutOfSession.cpp
jpk68: That's not really the point; the bug is that `h[24..32]` was uninitialized memory, which ends up in a header sent to peers. It's leaking stack contents over the wire. `RAND_bytes` is already used in `SendHolePunch`.

orignal: Then memory leak is wrong name. I agree about possible data leak hence memset should be fine.

wipedlifepotato:
Can you show your valgrind logs? where valgrind found the potential leak

libi2pd/Destination.cpp
jpk68: This is in `Stop`, not the destructor, so the destination will be kept alive too long. Canceling it here releases the reference like the other timers above it do.

orignal:
> Can you show your valgrind logs? where valgrind found the potential leak

No, I want your own analysis, what causes valgrind complaints (usually it always complains)

wipedlifepotato:
LeaseSetDestination::LeaseSetDestination (boost::asio::io_context& service,
		bool isPublic, const i2p::util::Mapping * params):
		m_Service (service), m_IsPublic (isPublic), m_PublishReplyToken (0),
		m_LastSubmissionTime (0), m_PublishConfirmationTimer (m_Service),
		m_PublishVerificationTimer (m_Service), m_PublishDelayTimer (m_Service), m_CleanupTimer (m_Service),
		m_LeaseSetType (DEFAULT_LEASESET_TYPE), m_AuthType (i2p::data::ENCRYPTED_LEASESET_AUTH_TYPE_NONE)
	{

wipedlifepotato:
m_PublishVerificationTimer in memory ~ m_PublishDelayTimer ~ m_Service? if yes then its already launch method stop? I confused

wipedlifepotato:
https://github.com/PurpleI2P/i2pd/blob/openssl/libi2pd/SSU2OutOfSession.cpp#L173

In stack is always will be zero
		uint8_t h[32], payload[SSU2_MAX_PACKET_SIZE];
...
		memset (h + 24, 0, 8); // header token

orignal:
> In stack is always will be zero

No, unless you write = {0};

wipedlifepotato:
```
#include<iostream>

int main(void)
{
        unsigned int h[32];
        for(auto i=31;i--;)
        {
                printf("%d\n", h[i]);
        }
}
```

Sure, yes.. ok

wipedlifepotato:
If this static var or [32]{} then will be zeros. I found in old literature
that sometimes even need set for any variable to zero
Like int I=0; but found these that will be zero. Before think that all vars
that not created thought new/malloc is empty. If no so,  would be good to
check all C arrays


вс, 26 июл. 2026 г., 04:55 orignal ***@***.***>:

PR #2456, opened 2026-07-30
chore: fix memory safety issues

jpk68:
- Fixes some incorrect blinding buffer sizes, causing a stack overflow
- Fixes a potential heap overflow in the `I2PControl` response writer
- Mitigates an integer underflow when handling datagrams

### DIFF
--- a/daemon/I2PControl.cpp
+++ b/daemon/I2PControl.cpp
@@ -286,7 +286,7 @@ namespace client
 					response << "{\"code\":-32601,\"message\":\"Method not found\"},";
 					response << "\"jsonrpc\":\"2.0\"}\n";
 				}
-				SendResponse (socket, buf, response, isHtml);
+				SendResponse (socket, response, isHtml);
 			}
 			catch (std::exception& ex)
 			{
@@ -295,7 +295,7 @@ namespace client
 				response << "{\"id\":null,\"error\":";
 				response << "{\"code\":-32700,\"message\":\"" << ex.what () << "\"},";
 				response << "\"jsonrpc\":\"2.0\"}\n";
-				SendResponse (socket, buf, response, isHtml);
+				SendResponse (socket, response, isHtml);
 			}
 			catch (...)
 			{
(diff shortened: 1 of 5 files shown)


libi2pd/Blinding.cpp
```diff
@@ -310,7 +310,7 @@ namespace data
 	i2p::data::IdentHash BlindedPublicKey::GetStoreHash (const char * date) const
 	{
 		i2p::data::IdentHash hash;
-		uint8_t blinded[128];
+		uint8_t blinded[i2p::crypto::ECDSAP521_KEY_LENGTH]; // 132, largest possible blinded public key (ECDSA-P521)
```
orignal: ECDSA-521 is not supposed to be used for blinding.
Only EdSA or RedDSA for now.
ECDSA-256 code should be removed too.
Please assume EdDSA.

libi2pd/LeaseSet.cpp
```diff
@@ -1001,7 +1002,7 @@ namespace data
 		auto timestamp = i2p::util::GetSecondsSinceEpoch ();
 		char date[9];
 		i2p::util::GetDateString (timestamp, date);
-		uint8_t blindedPriv[64], blindedPub[128]; // 64 and 128 max
+		uint8_t blindedPriv[i2p::crypto::ECDSAP521_KEY_LENGTH/2], blindedPub[i2p::crypto::ECDSAP521_KEY_LENGTH]; // 66 and 132 max (ECDSA-P521)
```
orignal: should be 32 and 64

daemon/I2PControl.cpp
```diff
@@ -306,28 +306,27 @@ namespace client
 
 	template<typename ssl_socket>
 	void I2PControlService::SendResponse (std::shared_ptr<ssl_socket> socket,
-		std::shared_ptr<I2PControlBuffer> buf, std::ostringstream& response, bool isHtml)
```
orignal, line 309: What is the purpose of this change?

jpk68: To fix the heap buffer overflow; `response` could exceed the buffer, so it now writes into an `std::ostringstream` instead.

libi2pd/LeaseSet.cpp
jpk68: Fixed.

libi2pd/Blinding.cpp
jpk68: You mean I should remove the ECDSA code from this file altogether?

orignal: Or just check if signature type EdDSA or RedDSA. Otherwise produce error. I will remove ECDSA blinding code later.

PR #2458, opened 2026-08-03
fix: macOS crash (cancel before close) + SOCKS5 IPv4 outproxy

bimawa:
### 1. Fix macOS crash: cancel pending socket ops before close
On macOS (kqueue), boost::asio throws an exception from pending async_write/async_read handlers after the socket is closed. This exception reaches io_service::run() as an uncaught exception and terminates the process (write_some: Bad file descriptor).

Call m_Socket.cancel() before shutdown/close in NTCP2, SAM, I2PTunnel, and BOB so pending operations complete with operation_aborted on a live socket.

### 2. Support IPv4 address type in SOCKS5 outproxy mode
When outproxy.enabled=true, i2pd rejects SOCKS5 CONNECT with IPv4 address type (0x01) returning rep=8. Many SOCKS5 clients (Telegram, browsers) send IPv4 CONNECT. Added ADDR_IPV4 to allowed types and conversion to DNS string for downstream forwarding.

Fixes: #TBD

### DIFF
--- a/libi2pd/NTCP2.cpp
+++ b/libi2pd/NTCP2.cpp
@@ -383,6 +383,7 @@ namespace transport
 			m_IsTerminated = true;
 			m_IsEstablished = false;
 			boost::system::error_code ec;
+			m_Socket.cancel ();
 			m_Socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
 			if (ec)
 				LogPrint (eLogDebug, "NTCP2: Couldn't shutdown socket: ", ec.message ());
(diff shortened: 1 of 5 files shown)


orignal:
>Call m_Socket.cancel() before shutdown/close in NTCP2, SAM, I2PTunnel, and BOB so pending operations complete with operation_aborted on a live socket.

Could you please refer to the boost specs/examples with this requirement?
socket::close should be enough to cancel all pending async operations.

orignal:
Please merge into openssl branch rather than master. ipv4 issue was fixed a while ago.

bimawa:
Thanks for info: here new version #2459

PR #2459, opened 2026-08-03
fix: cancel pending socket ops before close to prevent macOS crash

bimawa:
## Problem

On macOS (kqueue), boost::asio throws an exception from pending async_write/async_read handlers after the socket is closed. This reaches io_service::run() as an uncaught exception and terminates the process.

## Fix

Call m_Socket.cancel() before shutdown/close in NTCP2, SAM, I2PTunnel, and BOB so pending operations complete with operation_aborted on a live socket.

This is a kqueue-specific issue. On Linux (epoll), close() alone cancels pending operations correctly. The fix is defensive on Linux and required on macOS.

Boost docs reference:
https://www.boost.org/doc/libs/1_84_0/doc/html/boost_asio/reference/basic_stream_socket/close/overload1.html
> Any asynchronous send, receive or connect operations will be cancelled immediately,
> and will complete with the boost::asio::error::operation_aborted error.

This works on Linux. On macOS kqueue, the kernel completes the operation after the fd is closed, and boost::asio's reactor throws instead of delivering to the handler.

### DIFF
--- a/libi2pd/NTCP2.cpp
+++ b/libi2pd/NTCP2.cpp
@@ -574,6 +574,9 @@ namespace transport
 		{
 			m_IsEstablished = false;
 			boost::system::error_code ec;
+			#ifdef MAC_OSX
+				m_Socket.cancel ();
+			#endif
 			m_Socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
 			if (ec)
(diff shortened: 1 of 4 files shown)


orignal:
The please cancel sockets under
#if defined(MAC_OSX) or similar macro
if you believe that's the reason of crash.

bimawa:
ok thanks. Yes after that fix i2pd crashes end.

PR #2462, opened 2026-08-04
libi2pd: reset config state on initialization

### LINKED ISSUE #2345: libi2pd: Cannot call InitI2P twice
the precise error is
```
args: option '--datadir=/tmp/path_to_da is ambiguous and matches different versions of '--datadir'
```
it is raised at `i2p::config::ParseCmdline()`
I think it is because config::m_optionsDesc is filled twice and is never cleaned and does not seem to be cleanable by design. After the option gets 2 descriptions as a result of adding the same options another time, at parsing time this error occurs.

If reinitializing I2P gracefully is not an option, then I want to know: in our project we sometimes need to drop everything and redo all I2P bootstrap, how would one go around it in the most currently supported way?
It is also arising in unit testing when one has several tests in a test suite, because this usually involves reinitialization as a part of test fixture and we cannot clean globals between tests without splitting into different binaries for each test.

### Minimal reproduction:

https://gitlab.com/Houkime/i2p-restart-minimal

iasds:
Reset the config parser state before registering options.

Add a regression test for parsing command-line options after configuration reinitialization.

Fixes #2345.

### DIFF
--- a/libi2pd/Config.cpp
+++ b/libi2pd/Config.cpp
@@ -10,6 +10,7 @@
 #include <iostream>
 #include <fstream>
 #include <map>
+#include <memory>
 #include <string>
 #include <boost/program_options/cmdline.hpp>
 #include <boost/program_options/options_description.hpp>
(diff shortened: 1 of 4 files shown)


orignal:
options_description m_OptionsDesc;  is supposed to work as a constant structure and being filled only once.
Instead making it unique_ptr add a bool flag if it's intialized already and check and set it in Init.

PR #2469, opened 2026-08-11
Fix compilation when WIN32_APP is not defined

parsebulb:


parsebulb:
I'll make a small improvement

PR #2493, opened 2026-08-18
chore: fix memory safety issues

jpk68:


libi2pd/Blinding.cpp
```diff
@@ -183,7 +183,7 @@ namespace data
 
 	i2p::data::IdentHash BlindedPublicKey::GetStoreHash (const char * date) const
 	{
-		i2p::data::IdentHash hash;
+		i2p::data::IdentHash hash{};
```
orignal, line 198: Doesn't make any difference.

libi2pd_client/SAM.cpp
```diff
@@ -779,7 +779,7 @@ namespace client
 		auto res = std::from_chars(sizeStr.data(), sizeStr.data() + sizeStr.size(), size);
 		if (res.ec != std::errc()) size = 0;
 		size_t offset = data - buf;
-		if (offset + size <= len)
+		if (offset <= len && size <= len - offset)
```
orignal, line 782: why is this different than before?

jpk68: Checking `offset <= len` first means that `len - offset` can't underflow.

orignal: That why we don't calculate len - offset, but offset + size instead. Why do you need len - offset?

jpk68: Because `size` is attacker-controlled and unbounded before this check, so it can be made to wrap around. This change ensures that cannot happen.

For example, if `offset` is 10 and `len` is 1000, sending `SIZE=18446744073709551611` makes `offset + size` overflow to 5, even though the actual size is much larger.

orignal: No, size comes from a client app, not from network.  Bad client app can do many other bad things anyway.

PR #2511, opened 2026-08-23
torrents: fix path traversal, undefined behaviour

jpk68:


orignal:
Don't fix torrenting code for now, it's not ready yet.

jpk68:
> Don't fix torrenting code for now, it's not ready yet.

Fine, but please keep in mind there is a serious and currently exploitable vulnerability in the code that should be fixed.

PR #2512, opened 2026-08-23
http, i2pcontrol: use thread-safe snapshots for session/tunnel lists

jpk68:


libi2pd_client/I2CP.cpp
```diff
@@ -183,7 +188,7 @@ namespace client
 		}
 		m_ReadinessCheckTimer.cancel ();
 		auto pool = GetTunnelPool ();
-		if (!pool || pool->GetOutboundTunnels ().empty ())
+		if (!pool || pool->GetOutboundTunnelsList ().empty ())
```
orignal: You don't need to obtain full copy of tunnels. You only need to know is there are any. Please add method like HasInboundTunnels/HasOutboundTunnels to TunnelPool and check for empty there under mutex.

PR #2552, opened 2026-09-13
123

wipedlifepotato:


wipedlifepotato:
root@comp:/home/user/mi2pd# cp contrib/webconsole/* /var/lib/i2pd/webconsole/
root@comp:/home/user/mi2pd# 
image

PR #2557, opened 2026-09-16
fix: missing checks and limits

jpk68:


libi2pd/Reseed.cpp
```diff
@@ -270,7 +270,7 @@ namespace data
 			if (it != m_SigningKeys.end ())
 			{
 				// TODO: implement all signature types
-				if (signatureType == SIGNING_KEY_TYPE_RSA_SHA512_4096)
+				if (signatureType == SIGNING_KEY_TYPE_RSA_SHA512_4096 && signatureLength >= 64)
```
orignal: why 64?

jpk68: It's the length of a SHA-512 digest.

orignal: Could you please use constant  like  SHA512_DIGEST_LENGTH from openssl instead magic number?

